PR #16, opened 2022-11-10
Nex 6 storybook

ekondareva:
*Link to ticket: [NEX-6](https://wunder.atlassian.net/jira/software/c/projects/NEX/boards/389?modal=detail&selectedIssue=NEX-6)*

- Storybook is added to the project
- Articles stories are added as an example

*How to test:*
1. cd /next
2. run npm run storybook
3. check Articles stories


vermario (commented):
I tried this, and storybook starts correctly and I can see the (unstyled) article and aricle teaser.

Question: are we ok with using npm only? If so I suggest to remove yarn.lock

ekondareva:
> I tried this, and storybook starts correctly and I can see the (unstyled) article and aricle teaser.
> 
> Question: are we ok with using npm only? If so I suggest to remove yarn.lock

We agreed in the very beginning on using npm. I removed yarn.lock

vermario (approved):
I think this will work as a base for storybook, nice work :)

PR #18, opened 2022-11-15
NEX-16: add preview to the mostly automated setup of the project locally

vermario:
This branch adds instructions and automation to set up preview mode when kicking off a project using this starter, using a combination of a drupal recipe and a custom module that provides a drush command to set some parts up.

Following the instructions[ on the next-drupal website](https://next-drupal.org/learn/preview-mode), these are the steps needed: 

1. https://next-drupal.org/learn/preview-mode/create-site > added the site entity to the recipe
2. https://next-drupal.org/learn/preview-mode/create-oauth-client
   * role with permission -> added role and permissions to the recipe
   * key generation -> added lando script that generates the keys in the correct directory
   * creation of a user and the consumer connected to that user -> added a custom module with a drush command that creates the user, a random secret and the needed consumer, gives out the lines that need to be added to .env.local in next
3. Configure preview routes -> included in the starter we are using
4. https://next-drupal.org/learn/preview-mode/configure-content-types
   * this is done with configuration, so I have added this to the recipe.
   
   
To test: 

* follow the updated README from scratch, and you should be able to preview content, including unpublished content from within drupal.

(This handles the local development, we will still need to figure out how to make this work in silta etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,10 @@ Follow these steps to get started:
 1. `cd next`
 2. `npm install`
 3. `cp .env.example .env.local`
+4. back in the drupal directory, execute `lando drush wunder_next:setup-user-and-consumer`
+5. copy the lines that the command outputs into your `.env.local` file in the next directory.
 4. `npm run dev`
 5. visit `localhost:3000` and you should see your content displayed by the frontend.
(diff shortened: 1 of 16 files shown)


ekondareva (approved):
Great job!

PR #19, opened 2022-11-15
Nex 9  - translation support

ekondareva:
*Link to ticket: [NEX-9](https://wunder.atlassian.net/jira/software/c/projects/NEX/boards/389?modal=detail&selectedIssue=NEX-9)*

- Recipe wunder_languages added. It adds basic language setup for the site. 
- Custom module wunder_translations with the command added. It allows us to enable translations for Article and Basic page content types.
- README instructions are updated to set up the project
- Next (frontend): locale is added. Translated versions of articles are thus shown.

*How to test:*
1. Start from scratch and follow the instructions from an updated README file of the project.
2. Create Articles and translate into fi/sv
3. Open Next Preview / localhost:3000 and confirm you can see the translated versions of the articles.

<add silta url here>


vermario:
@ekondareva  I have added to this pr:

* next-i18next + configuration for the three languages, translated all existing string interfaces (expecially for swedish I used google translate... so it could be very wrong  :-) ) but it should show the point of how to add those
*  rudimentary language switcher

I also have ported some of the configuration from https://github.com/wunderio/nextjs-project, in terms of prettier, eslint and typescript configuration. I was wondering if it would make sense to rebuild the example on top of that, since it's supposed to have our current best practices... but this branch has already grown a lot...

To test this part: 
1. create articles and translate them to en, fi, sv
2. navigate to the site, use the language switcher to get to the list of contents
3. while on a translated article, use the language switcher to change the language version


I looked into having the default language with the /en path, but apparently the middleware solution presented in the docs works only on vercel hosting, and doing it without needs a lot of work (There's an example in the espoo project) so maybe we can leave it out of the template.

vermario:
I have added a couple of small things: 

* requiring the wunder_languages recipe from the wunder_next_setup recipe, so there's no need for one more step
* adjusted the language settings for article and page so that you can see the language selection when creating content

In my test this works nicely, so i think we can merge this.

vermario (approved):

PR #20, opened 2022-11-17
NEX-16 Smart language switcher.

ekondareva:
Smart language switcher

- Path translations fetched and passed to Lang Switcher component via context

*How to test:*
1. Create an Article or two and translate to Fi, Swe, En.
2. Go in localhost:3000/fi and go to your Fi article.
3. Switch langs in Lang switcher menu and make sure correct pages are rendered and no 404 happen
4. Confirm everything works for homepage
5. Confirm lang switching works if you start from English article first.


next/pages/_app.tsx
```diff
@@ -1,10 +1,21 @@
 import "styles/globals.css";
+import React from 'react';
 
 import { AppProps } from "next/app";
 import { appWithTranslation } from "next-i18next";
 
+const default_langs = {
```
vermario, line 7: I think this is a bit of a problem, because then in our instructions we would need to  tell people to come here and add the set of languages, which I think we should avoid.

vermario: for this react side of things, I think it would make sense to talk for example with @juhov or some other people that have actually implemented these kind of frontend sites, so we can _reuse_ their experience as well (since this is about reusability)

juhov: Unfortunately I don't have experience about language's as nelonen.fi is Finnish only.

There is some prior art from the next-drupal people in the Umami example. In their setup the languages among some other things are defined in a [site.config.js](https://github.com/chapter-three/next-drupal/blob/main/examples/example-umami/site.config.js) file.

ekondareva: The problem is slug part translation.
In Drupal out of the box you have the pattern, that alias follows the title of the article, thus Eng and Fi versions of the same article will have different URLs.
In Umami example, both Spanish and English URLs are identical, therefore they can use in their example
<Link href={asPath} locale={locale} passHref>
and we can't do that. Simply put locale in front of your [slug] doesn't help. We need to fetch translated paths and pass them to LangSwitcher.

vermario: @ekondareva I understand, we can't get their language switcher component, I propose we get their idea of that site.config file, and have there the settings for languages in one place.

We could have: 
* name of the language
* langcode
* homepage for the language ( "/" for english, "/sv" for swedish), etc

then we can feed that info info the query that you did , and maybe even use the language name in the language switcher? 

would this work you think?

vermario: I'm thinking... if the jsonapi response for the node could include the paths of its translation then we would not need to do additional queries in the frontend.

I will look into whether we can add a field like that.

vermario: ^ so, it seems it might be possible, but if we want to switch to graphql later we would need to do this work twice, so for now I propose: 
* let's try adopting that `site.config` file
* let's go with the multiple queries

juhov: My two cents: If the options for getting the translated paths are
A) customize JSON API in Drupal or
B) do one extra query in Next.js 

I favor B, because translatePathFromContext() is and documented plus it makes Next.js developer less dependent of Drupal. And not having custom stuff in Drupal is one less thing to maintain.

next/components/node--article.tsx
```diff
@@ -1,6 +1,7 @@
 import Image from "next/image";
 import { DrupalNode } from "next-drupal";
 import { useTranslation } from "next-i18next";
+import React, { useEffect, useState } from 'react';
```
vermario, line 4: @ekondareva  this looks like a leftover from some test?

vermario (approved):
This works correctly also in preview mode. I will merge it

PR #29, opened 2022-12-09
NEX-26: Add revalidation feature

vermario:
*Link to ticket: [NEX-26](https://wunder.atlassian.net/browse/NEX-26)*

This PR adds [incremental static regeneration](https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration) and [On demand revalidation](https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration#on-demand-revalidation) to our setup.

To make this work, this PR adds: 

* revalidate property for the index and slug pages
* telling drupal to use the lando proxied path for frontend "https://frontend.lndo.site" instead of the localhost:3000  (so this the url that will be used for revalidation and also for iframes inside the site
* adding the new env var "DRUPAL_REVALIDATE_SECRET"
* setting content types's property for revalidation

### A note on revalidation

This mechanism is pretty rudimentary compared to drupal's cache clearing. For example, I have set the path of the frontpage (for all languages) as a path to be revalidated for article pages. But, any content can be added to the menu, which is then shown on all pages: there is no way (that could find) to revalidate the whole site when a menu item changes, for example. So I have set the revalidate option for all pages, so after 60 seconds they will be regenerated again (and then served with the stale-while-revalidated method)

### Impact on silta (which is in progress) 

@tharna this branch adds a new env var called `DRUPAL_REVALIDATE_SECRET` that should be present and have the same value both for drupal and next in silta. Also there's this info on next.js's site about running this in kubernetes https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration#self-hosting-isr

*How to test:*
1. do the setup from scratch, *including lando rebuild, since we changed some of the lando setup in this branch* 
You can use this one line command to speed things up: 
```
lando rebuild -y && lando composer install && lando generate-oauth-keys && lando drush si --site-name="My great site name here" -y && lando install-recipe wunder_next_setup && lando drush wunder_next:setup-user-and-consumer
```
2. copy the secrets like always to the `.env.local` file
3.  to test revalidation, you have to create a production build and run the server in production mode [like they say on next.js' website](https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration#testing-on-demand-isr-during-development) so run: 
```
lando npm run build
lando  npm run start
```
4. create some content
5. visit the site at https://frontend.lndo.site, not in preview mode (you can open the site in a private window to be sure
6. create an article
7. see that it appears at the forntpage of the site, and you can navigate into it
8. update the article, you should be able to see the changes straight away if you reload the page in the browser
9. create a new article, the homepage should update right away.
> starting from this PR, we should not use localhost:3000 but instead https://frontend.lndo.site to access the site

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -69,7 +69,12 @@ services:
     overrides:
       environment:
         # URL OF THE FRONTEND SITE:
-        WUNDER_NEXT_FRONTEND_URL: http://localhost:3000
+        # This needs to match the proxy value for the node service
+        # so if you change it, remember to change it here as well.
+        WUNDER_NEXT_FRONTEND_URL: https://frontend.lndo.site
+        # This same var is set in the node service below, the value should match.
+        # This is only used in local development so it can be a not secure value:
+        DRUPAL_REVALIDATE_SECRET: mysecret
         HASH_SALT: notsosecurehashnotsosecurehashnotsosecurehash
         ENVIRONMENT_NAME: lando
         DB_NAME_DRUPAL: drupal9
@@ -82,6 +87,20 @@ services:
         PHP_IDE_CONFIG: serverName=appserver
         # PHPUnit settings. @see: .lando/phpunit.sh
         MINK_DRIVER_ARGS_WEBDRIVER: '["chrome", {"browserName":"chrome","chromeOptions":{"args":["--disable-gpu","--headless"]}}, "http://chrome:9515"]'
+  node:
+    type: node:16
+    ssl: true
+    sslExpose: true
+    overrides:
+      ports:
+        - "3000:3000"
+      environment:
+        # This same var is set in the appserver service above, the value should match.
+        # This is only used in local development so it can be a not secure value:
+        DRUPAL_REVALIDATE_SECRET: mysecret
+    build:
+      - "cd next && npm install"
+    scanner: false
   chrome:
     type: compose
     services:
@@ -126,14 +145,6 @@ services:
     type: mailhog
     hogfrom:
       - appserver
-  node:
-    type: node:16
-    overrides:
-      ports:
-        - "3000:3000"
-    build:
-      - "cd next && npm install"
-    scanner: false
 
 proxy:
   mailhog:
@@ -143,6 +154,9 @@ proxy:
   # kibana:
   #   - kibana.lndo.site:5601
   node:
+  # This value is used by the appserver service to connect to the frontend
+  # so make sure that this value and the WUNDER_NEXT_FRONTEND_URL env var are
+  # in sync.
     - frontend.lndo.site:3000
 
 events:
--- a/drupal/recipes/wunder_next_setup/config/next.next_entity_type_config.node.article.yml
+++ b/drupal/recipes/wunder_next_setup/config/next.next_entity_type_config.node.article.yml
@@ -7,7 +7,7 @@ configuration:
   sites:
     frontend: frontend
   field_name: {  }
-revalidator: ''
+revalidator: path
 revalidator_configuration:
-  revalidate_page: null
-  additional_paths: null
+  revalidate_page: true
+  additional_paths: "/\r\n/fi\r\n/en\r\n/sv"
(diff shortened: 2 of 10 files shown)


next/pages/[...slug].tsx
```diff
@@ -112,5 +112,6 @@ export async function getStaticProps(
         ["common"]
       )),
     },
+    revalidate: 60,
```
Comment (vermario) [inline next/pages/[...slug].tsx:115]: without this, the page would never be updated when running a production build. From now on, we should always test both with `lando npm run dev` and `lando npm run build && lando npm run start`

drupal/recipes/wunder_next_setup/config/next.next_entity_type_config.node.article.yml
```diff
@@ -7,7 +7,7 @@ configuration:
   sites:
     frontend: frontend
   field_name: {  }
-revalidator: ''
+revalidator: path
 revalidator_configuration:
-  revalidate_page: null
-  additional_paths: null
+  revalidate_page: true
+  additional_paths: "/\r\n/fi\r\n/en\r\n/sv"
```
vermario, line 13: this tells drupal to revalidate the frontpage path for all languages when an article is created (this because in our site at the moment the frontpage displays a list of articles)

This is a poor's man cache-tags kind of idea...

.lando.yml
```diff
@@ -82,6 +87,20 @@ services:
         PHP_IDE_CONFIG: serverName=appserver
         # PHPUnit settings. @see: .lando/phpunit.sh
         MINK_DRIVER_ARGS_WEBDRIVER: '["chrome", {"browserName":"chrome","chromeOptions":{"args":["--disable-gpu","--headless"]}}, "http://chrome:9515"]'
+  node:
+    type: node:16
+    ssl: true
```
vermario, line 92: this is needed to have the https://frontend.lndo.site using https, which is needed because otherwise the browser does not allow to have a http iframe inside an https site.

vermario: (this means they will need to accept the warning, or trust lando's ca authority)

juhov:
Comment/question about IRS and revalidation in Silta environments:

Given the assumptions:
1. There is 1+n node processes (with their own file system) running at any given moment in the cluster.
2. The ISR cache is basically just a html file in the file system.

So in practice when Drupal makes a request to the revalidation endpoint, only one of the pods(?) will receive it and remove the file and the caches will drift apart, right?

Ping @vermario

vermario:
> Comment/question about IRS and revalidation in Silta environments:
> 
> Given the assumptions:
> 
>     1. There is 1+n node processes (with their own file system) running at any given moment in the cluster.
> 
>     2. The ISR cache is basically just a html file in the file system.
> 
> 
> So in practice when Drupal makes a request to the revalidation endpoint, only one of the pods(?) will receive it and remove the file and the caches will drift apart, right?
> 
> Ping @vermario

I think yes that seems to be the risk.

Reading https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration#self-hosting-isr , it seems to suggest that the default is storing the data in memory, and instead they suggest to store it in the filesystem, and make that filesystem shared for all nodes.

juhov:
Should we create a ticket for the shared file system? I think it's the best option even thought there is this 👇  race condition risk mentioned in the Next.js docs.

>Note: You might need to consider a race condition between multiple pods trying to update the cache at the same time, depending on how your shared mount is configured.

vermario:
@juhov we have an existing ticket for the silta setup, I will add a comment there https://wunder.atlassian.net/browse/NEX-11

juhov (approved):
I'm not qualified to approve the Drupal code changes, but everything works by testing trough the UI 👍

I tested all the steps mentioned in "How to test" in PR's description. I also tested without IRS (by removing all the `revalidate: 60`'s) and the on demand revalidation cleared Next.js cache correctly.

juhov:
Note about revalidation: One could argue that having the default 60s **ISR** time for everything is sort of defeating the point of **On demand revalidation**, but that should be optimized case-by-case for each project - and each route.

Those "menu on every page" or "latest articles list on every page" could in some cases be resolved by moving them to be loaded in the frontend and doing the "caching" in a Next.js API route + Cache-Control headers that proxies the data from Drupal.

saqib053 (approved):

PR #31, opened 2022-12-23
NEX-12: elasticsearch

vermario:
This PR adds elasticsearch search functionality to the site, both in the drupal backend and in the next frontend. It's a big one! :) 

### Backend

* Added elasticsearch to lando
* added elasticsearch helper (7.0 version)
* added custom module with: 
  * index definitions
  * normalizer
  * proxy to make actual searches
  
### Frontend

The frontend work is based on the existing https://github.com/wunderio/nextjs-elastic-search-ui-example effort

* Added latest version of https://github.com/elastic/search-ui
* Overridden ui components to be able to translate the ui strings, and use tailwind css to make the ui consistent, and show how the overriding can be done
* added facets and query matching the schema and content 
* added search page and search bar in the navbar in all pages
* added wunder colors

The "frontend of the frontend" implementation (classes, search bar in navbar etc) should be improved particularly in terms of responsiveness, but I propose to make that an independent jira story.

### How to test

1. set up the project from scratch, by following carefully the README file, which now includes also a step to create elasticsearch indices. (run lando rebuild if you have the project running already, since this adds elasticsearch to lando)
2. log into the site, create content and translate it. Add tags to the article content type
3. access `https://frontend.lndo.site/search`
* click the search button and see all results
* search for a keyword and see results
* input a search result that won't find any result and see the message about no search results found

Switch to a different language and navigate to /search

### Known limitations

* when clicking on the language switcher while on the /search page, you are taken to the frontpage instead of the translated search page, this is due to how the language switcher works. For now we decided to go with this.

### Screenshots

image


next/tailwind.config.js
```diff
@@ -5,8 +5,25 @@ module.exports = {
     "./components/**/*.{js,ts,jsx,tsx}",
   ],
   theme: {
-    extend: {},
+    extend: {
+      colors: {
+        // Generated with https://uicolors.app/
+        wunderpurple: {
```
vermario, line 11: This gives us automatically tailwind classes that we can use to show our colors in the site

next/pages/search.tsx
```diff
@@ -0,0 +1,148 @@
+import "@elastic/react-search-ui-views/lib/styles/styles.css";
```
vermario, line 1: This is the main file for the search ui

next/pages/search.tsx
```diff
@@ -0,0 +1,148 @@
...
+            >
+              {({ wasSearched, results }) => (
+                <ErrorBoundary>
+                  <div className="search-ui">
+                    <SearchBox
+                      searchAsYouType={false}
+                      shouldClearFilters={false}
+                      // Here we specify out own custom
+                      // component to render the search bar:
+                      inputView={SearchBoxInput}
```
vermario, line 82: We are providing our custom component to render the view of the search bar. We have repeated this for all visible components

.lando.yml
```diff
@@ -106,41 +106,41 @@ services:
...
-  #   services:
-  #     image: "docker.elastic.co/kibana/kibana:7.17.0"
-  #     command: "/bin/tini -- /usr/local/bin/kibana-docker"
-  #     user: kibana
-  #     ports:
-  #       - "5601:5601"
+  elasticsearch:
+   type: compose
+   services:
+     image: "docker.elastic.co/elasticsearch/elasticsearch:7.17.0"
```
misterjoonas, line 112: Why elasticsearch 8?

vermario: I had the same idea, I tried initially to use elasticsearch 8 but it didn't work locally (can't remember the exact issue), and apparently it has not been used in projects yet, so I didn't want to expand the scope too much.

(discussion with Janne about this https://wunder.slack.com/archives/C049GMMBEAC/p1671794114416579 )

vermario: Maybe we could have it as a followup task?

drupal/web/modules/custom/wunder_search/src/Plugin/Normalizer/NodeNormalizer.php
```diff
@@ -0,0 +1,276 @@
...
+   * @param \Drupal\Core\Entity\ContentEntityBase $entity
+   *   The entity.
+   * @param array $context
+   *   The context (used to get the language).
+   *
+   * @return string
+   *   The string to use as content label.
+   */
+  protected function getContentTypeLabel(ContentEntityBase $entity, array $context) {
+    $bundle_labels_map = [
```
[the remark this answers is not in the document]
misterjoonas: Something like https://drupal.stackexchange.com/a/187983

misterjoonas (approved):
Looking good and works!

drupal/web/modules/custom/wunder_search/src/Plugin/Normalizer/NodeNormalizer.php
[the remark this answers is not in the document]
vermario: yep, that works.

This comes from a project where they wanted to customize the name of the content type in search results (but not in the admin interface) and translate it into various languages... maybe the simpler version is enough for this template.... I will simplify it.


(unfortunately we cannot add translated bundle names via recipe, because translated configuration is not supported)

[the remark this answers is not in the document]
vermario: (changed in https://github.com/wunderio/next4drupal-project/pull/31/commits/f9778462e30aabadcf9b2cfb341a976465cbfaa6 )

PR #38, opened 2023-02-07
Various NextJS improvements

joshua-scott:


vermario (approved):
I built the site locally, and it works.

Warnings about eslint next are gone.

I think we can merge this.

PR #40, opened 2023-02-09
Add back a patch for the jsonapi_menu_items module that adds the langcode

vermario:
I have added a new version of the patch on d.org: https://www.drupal.org/project/jsonapi_menu_items/issues/3192576#comment-14914978

This should add back the langcode when getting menu items from jsonapi.

To test

1. delete the web/modules/custom folder
2. run `lando composer install` so the patch is applied
3. test the response of menus


joshua-scott (approved):
It works - menu links are back!

CleanShot 2023-02-09 at 17 22 55@2x

PR #41, opened 2023-02-10
NEX-27: Improvements in UX for the drupal side

vermario:
https://wunder.atlassian.net/browse/NEX-27

This branch includes these changes for the backend site: 

1. adding the `require_login` module, so the user is presented with a login form before accessing the site. A list of paths are excluded from this mechanism, because they are meant to be accessed by the frontend. For example jsonapi paths, or our elasticsearch proxy path.

2. setting the "claro" admin theme as the "default" theme, and the `admin/content` page as the frontpage, to make it clearer that the backend is used only for content creation / administration.

### To test: 

1. execute the usual mega command: `lando composer install && lando generate-oauth-keys && lando drush si --site-name="My great site name here" -y && lando install-recipe wunder_next_setup && lando drush wunder_next:setup-user-and-consumer && lando drush eshd -y && lando drush eshs`
2. complete the setup copying the access data to `next/.env.local` as instructed
3. verify that visiting the backend you see a login form
4. login using username `admin` and the password that was shown to you when running the drush si command
5. navigate to the frontpage of the backend, and confirm that it's the list of content
6. create some content and confirm that it shows up correctly in the frontend 

[Screencast from 10-02-23 12:50:50.webm](https://user-images.githubusercontent.com/185412/218074276-0b63fe73-88b6-47d0-a620-be84ae1e279a.webm)


drupal/web/modules/custom/wunder_next/wunder_next.install
```diff
@@ -0,0 +1,30 @@
...
+ * @file
+ * Install, update and uninstall functions for the wunder_next module.
+ */
+
+/**
+ * Implements hook_install().
+ */
+function wunder_next_install() {
+
+  // Currently these configs cannot be set via recipes, so we
```
vermario, line 13: This is a bit funky, but no other way at the moment...

vermario: Drupal slack to the rescue! I have replaced this with the correct way of updating existing configuration.

joshua-scott (approved):
Testing and working great 👍

PR #43, opened 2023-02-10
NEX-33: Responsive layout and other frontend improvements

joshua-scott:
Implements https://wunder.atlassian.net/browse/NEX-33 but as usual I got carried away and did a bunch of other unrelated stuff...

## Highlights

- Improved responsive layout for the header and footer, and a bit for `main` (but still more to figure out when we've put some more content there)
- Add [tailwind prettier plugin](https://tailwindcss.com/blog/automatic-class-sorting-with-prettier) to sort classes and run it
- Support SVGs
- Simplify/improve language switcher
- Upgrade storybook to use [v7 with built-in Next.js support etc](https://storybook.js.org/blog/integrate-nextjs-and-storybook-automatically). As we start to add components, the plan is to use storybook a bit more too. But for now it's very much the default setup.

## How to test

1. From the /next directory,`npm install && npm run build && lando run start`; does everything still work as expected?
2. `npm run storybook`, http://localhost:6006/ should open and work to the default "Welcome to storybook" page


vermario (approved):
Works, and a very nice upgrade for the frontend :)

I have added two small commits:

1. removed a console.log that I had left somewhere
2. told lando to map port 6006 (which is used by storybook) so we can also use `lando npm storybook) and then access it via localhost:6006

PR #44, opened 2023-02-13
NEX-39: Add frontpage handling

vermario:
This branch adds the possibility for the content editor to decide what is displayed on the frontpage of the site for each of the translations.

### How it works

In the backend: 
* there is now a new content type called "frontpage". Paragraphs can be added to the frontpage and it can be translated
* there is a constraint that prevents editors to add more than one frontpage translation per language
* the path of the frontpage is set via pathauto
* the frontend node is not indexed in elasticsearch (this PR also fixes an issue where items could not be deleted from elasticsearch)

On the frontend:
* the `index.tsx` contains a query to get the frontpage for the language. if found, its contents (paragraphs) are rendered at the top of the page, before the fixed listing that shows the latest articles.
* because the frontpage also has its own url in drupal (something like /en/frontpage-en for english for example), the [...slug.tsx] page now contains code that will issue a redirect to the root of the site if the path corresponds to a frontpage node. This way we are able to preview the frontpage when visiting the "view" tab for the frontpage node.

### How to test

* run the usual big command:  `lando composer install && lando generate-oauth-keys && lando drush si --site-name="My great site name here" -y && lando install-recipe wunder_next_setup && lando drush wunder_next:setup-user-and-consumer && lando drush eshd -y && lando drush eshs` and then copy the usual credentials to `.env.local` 
* create a frontpage node for english
* translate into finnish
* translate info swedish
* try to add a new frontpage node -> you will see a message warning you that it already exists
* on the content list, click on the frontpage node you created in all languages, check that the contents are displayed in the preview
* visit the frontend site and see your contents displayed 

### Screenshots

Creating a frontpage: 
image

Trying to add a second frontpage for a language: 
image

Viewing the preview of the frontpage, where the iframe has been redirected to /en
image

### DIFF
--- a/drupal/recipes/wunder_landing_pages/README.md
+++ b/drupal/recipes/wunder_landing_pages/README.md
@@ -5,8 +5,11 @@ When installed, this recipe will create:
 * a "landing page" content type with:
   * field definitions
   * pathauto settings
-  * text paragraph
-  * image paragraph
+  * paragraphs
+* a "frontpage" content type with:
+  * field definitions
+  * pathauto settings
+  * paragraphs
 
 ### Manual Installation instructions
 
--- a/drupal/recipes/wunder_landing_pages/config/core.base_field_override.node.frontpage.promote.yml
+++ b/drupal/recipes/wunder_landing_pages/config/core.base_field_override.node.frontpage.promote.yml
@@ -1,0 +1,22 @@
+uuid: 0cfed645-1ed6-4dca-9fea-faf7b0096c58
+langcode: en
+status: true
+dependencies:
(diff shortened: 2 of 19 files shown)


joshua-scott (approved):
Tested and it's working nicely, including in preview mode.

One small thing to consider about the title; I don't know if this is difficult to do, but might be worthwhile if it's a quick one? Feel free to merge if not.

CleanShot 2023-02-14 at 12 27 32@2x

joshua-scott:
Yep, that's true - this title could be useful for that. Something to keep in mind for later, then 👍

PR #45, opened 2023-02-13
NEX-28: Use component library tokens. Improve layout. Share common page props. Add 404/500 pages... etc

joshua-scott:
- Implements [NEX-28](https://wunder.atlassian.net/browse/NEX-28), which adds style tokens from the in-progress [Wunder component library](https://www.figma.com/file/i0RIoStoPOZfcqS80DLbkD/The-Component-Library). Some basic typography styling has been applied but there's definitely more to do in this area.
- Continuing from  [NEX-33](https://wunder.atlassian.net/browse/NEX-33), improves the Layout component to be added once to _app.tsx, instead of separately on every page. This approach prevents a re-render when switching between pages, and is the approach recommended by Next.js when [only one layout type is needed](https://nextjs.org/docs/basic-features/layouts#single-shared-layout-with-custom-app) (let's see if we need to change it later to support multiple layout types - maybe it'll be a good idea as it's quite likely to be required for most projects)
- Add 404 and 500 pages (`next build` would error without this, since the Layout expects the menu data to always be there)
- Adds fonts using the fancy new [NextJS google fonts approach](https://nextjs.org/docs/basic-features/font-optimization#with-tailwind-css)
- Create getCommonPageProps function and use it on all pages
- Add back the PreviewBanner
- Some random type improvements as I came across them
- Add type checking to eslint.

To test, `lando npm i && lando npm run dev`, and check several pages of the site (including 404). Do they still work and look decent with no errors? Also, does the preview banner show when you click the preview content in Drupal to open it in its own tab?

[NEX-28]: https://wunder.atlassian.net/browse/NEX-28?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NEX-33]: https://wunder.atlassian.net/browse/NEX-33?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vermario (approved):
I tested this locally by adding some content, running build for the site. Works nicely! let's merge it.

PR #49, opened 2023-02-20
NEX-43

vermario:
Included in this PR:

1. switch from the `default` installation profile to the `minimal` installation profile, which allows us to leave out all unneeded modules and configuration
2. remove the standard page content type (now that we don't _have_ to have it), and instead rename the landing_page content type to just page
3. smaller fixes:
  * Renamed the imported "landing page" to "Paragraph example"
  * fix the import of video paragraph in the english paragraph page
  * make sure the build does not fail if a frontpage node is not there (like when running the megacommand in the beginning)
  
  
### To test

The command has been updated in the readme: 

```
lando rebuild -y && lando composer install && lando generate-oauth-keys && lando drush si minimal -y && lando install-recipe wunder_next_setup && lando drush wunder_next:setup-user-and-consumer && lando drush eshd -y && lando drush eshs && lando npm i && lando npm run build && (lando npm run start&) && lando drush en wunder_democontent -y && lando drush mim --group=demo_content --execute-dependencies && lando drush uli
```


joshua-scott (approved):
Tested and it's working nicely 👍

PR #50, opened 2023-02-22
NEX-20: Add metatag support

vermario:
This PR adds metatag support to the site, and also adds some adjustments for the search and the backend block layout.

On the backend:

* added a field called "excerpt" to all content types
* added to the search index
* added the field to migration
* improved the article import data with chatgpt-created generic drupal throwaway article content
* added the metatag module to the recipe (including a patch to add the metatag output to jsonapi)

On the frontend:

* switched to using the excerpt in search results
* added translated site name and site description to i18n json files
* added an env var: NEXT_PUBLIC_FRONTEND_URL: https://frontend.lndo.site to the lando setup (*rebuild necessary*)
* added the <meta> component that combines fixed metatags and metatags coming from the nodes.
* added the component in all pages, both created by drupal and not (search page, error pages)

Seems to work!

### Testing instructions

1. Run the megacommand. This time a rebuild is needed because we added one more env var for the node environment.
2. visit pages and see the metatags. For pages that are built by nodes, it will have the title + translated site name as the title, for other pages (like search) it should use the default values for that language. For arcles pages, the image metatag will point to the actual image, for nodes with no images and for hardcoded pages (search etc) the image should point to a default image that I added in the public folder.


drupal/config/sync/language/fi/views.view.media.yml
```diff
@@ -1,18 +0,0 @@
-display:
```
vermario, line 1: These config files were committed by mistake, let's remove them.

next/.env.example
```diff
@@ -1,8 +1,16 @@
+# When developing locally, you should use node indside lando, which
```
vermario, line 1: this should make it clearer that you should use lando instead of .env files + localhost

next/components/meta.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const languagePathFragment =
+    router.locale === router.defaultLocale ? "" : `/${router.locale}`;
+
+  const data = {
+    title: getTag("title")?.content ?? title,
+    description: getTag("description")?.content ?? t("meta-site-description"),
+    canonical: `${process.env.NEXT_PUBLIC_FRONTEND_URL}${languagePathFragment}${
+      router.asPath !== "/" ? router.asPath : ""
+    }`,
+    imageSrc: getTag("image_src", "rel")?.href,
```
vermario, line 48: do we take this out, or try to see how to provide that image?

next/components/meta.tsx
```diff
@@ -0,0 +1,82 @@
...
+        images: [
+          {
+            url: data.imageSrc ?? "",
+            alt: title,
+          },
+        ],
+      }}
+      additionalMetaTags={[
+        {
+          httpEquiv: "x-ua-compatible",
```
vermario, line 72: these seemed generic enough that we might want to keep them? to show how to add other metatags?

next/lib/get-params.ts
```diff
@@ -20,6 +20,7 @@ export function getNodePageJsonApiParams(resourceType: ResourceType) {
         "field_content_elements",
         "path",
         "status",
+        "metatag",
```
vermario, line 23: this will need to be adapted in the zod branch (as the rest of changes in this file)

next/.env.example
joshua-scott: Good idea. Another option would be to remove this file entirely - is there a reason we want to keep it?

next/components/meta.tsx
joshua-scott: I'm not really sure about this. I'd lean towards leaving it out for now and adding it later if there's a need?

(I did search our GitHub https://github.com/search?l=JavaScript&q=org%3Awunderio+img_src&type=Code to see if there was a clear and tested way of doing it in Wunder, but didn't find anything.)

next/components/meta.tsx
joshua-scott: I agree, it's good to show as an example. But let's maybe drop the first "ie=edge" one (apparently it's not been necessary for while) and keep the second one? I don't have strong opinions about it though.

next/lib/get-params.ts
joshua-scott: Yep. I vote we merge this branch first and I can adapt mine after and handle the conflicts etc.

joshua-scott (approved):
Tested by checking out the metatags in the frontend and I can see that they're all there and changing as expected on different pages, languages etc. This is a great start, we can look into small improvements later if needed.

next/.env.example
vermario: good point, removed in 4b2f0fdb5aef796b232ed9d9661b8c59309c0af7

next/components/meta.tsx
vermario: I have added the necessary configuration in 1008b790ab754a6c19d6bc26548abddfdbafab9b for drupal to print the url of the field_image image file into the metatag. This works for articles only, since we don't have that field in other content types.

Next step would be maybe to provide a default image in the next frontend (public folder?) and print a link to that one in case the metatags from drupal don't include an image

next/components/meta.tsx
vermario: ok! removed the first one in 3231192c4bba8b7ce055260915da1d965695c8f9

next/components/meta.tsx
vermario: I added the fallback in f11837e67a19c0b8f6723ab59b8a9df90342e395

vermario:
@joshua-scott , please re-review this, I realised that we were passing an empty parameter for the metatags (the property in the resource is called `metatag`, not `metatags`

So we were only seeing the fallback and not the actual metatags at work. 

I also added the image and the fallbcak image.

joshua-scott (approved):
👍

PR #52, opened 2023-02-22
Use zod to validate and cleanup page data

joshua-scott:
Adds [Zod](https://zod.dev) to validate and cleanup data from the backend. 

By "validate", we mean that data is checked at runtime to ensure that the fetched data is of the expected shape and type. This can help catch bugs because if the data ever changes, we will notice it quickly during development.

By "cleanup", we mean that any data properties that are not listed in the `next/lib/zod/*.ts` types are discarded when its corresponding "validateAndCleanup[MyCoolResource]" function is called. Because we always call this function on the server side (in getStaticProps), we clean the data before it is passed to the clientside as page props. Therefore, the client downloads fewer bytes and the page is faster to load.

## Testing

To test, `lando npm run dev` and check that every page still loads as expected with no missing data. You could also try adding/removing/editing some content and checking it's still working.


vermario (approved):
To my eyes this works nicely. I will have to get used to zod going forward :-)

PR #56, opened 2023-03-02
Show excerpt on article pages

joshua-scott:
This PR shows field_excerpt on article pages. It does NOT add it to the article "teaser" list on the frontpage.

CleanShot 2023-03-02 at 11 30 38@2x

To test, run the usual `lando npm run dev` and navigate to an article page - you should see the excerpt if available.


vermario (approved):
nice!

PR #64, opened 2023-03-23
NEX-46: Dynamically import Accordion and Video paragraphs

joshua-scott:
https://wunder.atlassian.net/browse/NEX-46

This PR adds and example of using [Next.js Dynamic Imports](https://nextjs.org/docs/advanced-features/dynamic-import) to defer loading of the ParagraphAccordion and ParagraphVideo components until they're needed. It does seem to to have a (modest) effect - the "First load JS" on the homepage is 6kB less on the homepage where neither component is used. For this template, though, it's more about introducing the pattern, so developers are aware that this is the way to prevent the Paragraph component growing over time.

`next build` before and after:

pic

### DIFF
--- a/next/components/paragraph.tsx
+++ b/next/components/paragraph.tsx
@@ -1,10 +1,18 @@
+import dynamic from "next/dynamic";
+
 import { ParagraphImage } from "@/components/paragraph--image";
 import { ParagraphLinks } from "@/components/paragraph--links";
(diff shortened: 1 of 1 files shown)


vermario (commented):
I tested this locally, works nicely!

vermario (approved):

PR #69, opened 2023-04-19
Typesafe environment variables

joshua-scott:
This PR enforces typesafe env variables in the frontend using zod. The goal is that if an env var would be missing from somewhere, zod will error loudly at runtime and make it clear where the issue is.

If adding a new env variable that will be used on the frontend (e.g. to `.lando.yml`, or possibly to `silta*.yml`), also add it to the new file at `next/env.ts`. You can then import and use it with `import { env } from "@/env"` to get type safety and autocomplete.


vermario:
Let's add a bit of documentation in readme still

vermario:
@joshua-scott we were indeed missing two env vars! I added them to the silta setup, and now the build works :)

vermario (approved):
Let's merge!

PR #75, opened 2023-04-26
NEX-11:  Generate the oauth keys during deployment

vermario:
This small branch removes the hardcoded oauth keys, and instead uses the drush command provided by the simple_oauth module to generate them during deployment.


silta/silta.yml
```diff
@@ -12,14 +12,14 @@ php:
...
-  # This section is used in the template only, it will reinstall the site and apply
-  # all standard configuration at each commit. When creating a new project from this template
-  # you have to remove this section.
   postupgrade:
     afterCommand: |
       drush si minimal -y
       chmod +x /app/web/core/scripts/drupal
       cd /app/web && core/scripts/drupal recipe ../recipes/wunder_next_setup
+      drush cr
+      drush simple-oauth:generate-keys /app/oauth
```
tharna, line 21: This won't work as the keys are generated on a running container storing the keys only to that containers ephemeral storage instead of storing them to the actual container image so they will be lost after the container is recreated (also not present on drupal pod as this is being run on shell pod).

vermario: ah, that explains it :)

PR #81, opened 2023-05-03
NEX-61: Do not show preview buttons in content types edit form

vermario:
Jira: https://wunder.atlassian.net/browse/NEX-61

This PR hides the standard Drupal "preview" buttons in all existing content types, to prevent confusion with the next.js preview.


Zionknight-crypto (approved):
I have done a manual check on the feature env and code is straight forward. Great!

PR #82, opened 2023-05-04
Nex 62: a11y improvements

Zionknight-crypto:
[NEX62](https://wunder.atlassian.net/browse/NEX-62)

- Audit and improve accessibility

*How to test:*
1. Take feature link, login and run lighthouse analysis (in incognito) to view the accessibility score. 
2. I recommend front page, basic page and search page with results.

https://feature-nex-62-a11y-next.next4drupal-project.dev.wdr.io/


next/components/article-teaser.tsx
```diff
@@ -36,6 +36,7 @@ export function ArticleTeaser({ article }: ArticleTeaserProps) {
       <Link
         href={article.path.alias}
         className="absolute inset-0 cursor-pointer"
+        aria-label={`${t("article-link")}${" | "}${article.title}`}
       />
```
joshua-scott, line 40: I'm wondering if we need the "Article link | " prefix here? Screen readers etc would announce it as a link so maybe it's redundant.

Idea: This link could instead have aria-labelledby={titleId}, which we can pass to the h3 on line 20? Rather than duplicating info in the label again

next/components/header.tsx
```diff
@@ -52,8 +52,15 @@ function SearchLink() {
...
+      href="/search"
+      locale={locale}
+      className="hover:underline"
+      aria-label={t("search-link")}
+    >
+      <span className="sr-only sm:not-sr-only sm:mr-2 sm:inline">
+        {t("search")}
+      </span>
       <SearchIcon className="inline-block h-6 w-6" />
     </Link>
```
joshua-scott, line 65: Here, I feel like aria-label should be unnecessary because the child span (which is always visible to screen readers) already tells us what it does?

If the automated a11y software says it IS needed, can we again use aria-describedby to point to the span, instead of adding another label with its translation etc?

next/components/language-switcher.tsx
```diff
@@ -20,11 +21,17 @@ export function LanguageSwitcher() {
...
-        <span className="hidden sm:mr-2 sm:inline">
+      <button
+        type="button"
+        className="hover:underline"
+        onClick={toggle}
+        aria-label={t("language-switcher-button")}
+      >
+        <span className="sr-only sm:not-sr-only sm:mr-2 sm:inline">
           {languageLinks[locale].name}
         </span>
```
joshua-scott, line 36: (same question about aria-label duplication here)

next/components/main-menu/main-menu.components.tsx
```diff
@@ -177,6 +184,7 @@ export function MenuTrigger({ isTopLevel }: { isTopLevel?: boolean }) {
           ? "lg:ring-white lg:aria-expanded:bg-white lg:aria-expanded:text-primary-600"
           : "lg:aria-expanded:bg-primary-600 lg:aria-expanded:text-white lg:aria-expanded:ring-primary-600"
       )}
+      aria-label={`${t("show-submenu")}${" "}${t(parent)}`}
```
joshua-scott, line 187: This looks good, but we don't need to run `parent` through `t()` because it's already translated in Drupal side.

Proposed solution: in common.json, we can have `"show-submenu": "Show submenu for {{parent}}",`, and then we can use it like `aria-label={t("show-submenu", { parent })}` (see the translation `form-submission-intro-text` for another example)

next/components/search/search-result.tsx
```diff
@@ -4,21 +4,23 @@ import { Badge } from "@/wunder-component-library/badge";
...
-      <p className="mb-6 text-md text-scapaflow">{result.excerpt.raw}</p>
-      <Badge variant="info" size="sm">
-        {result.content_type.raw}
-      </Badge>
-    </a>
+    <li>
+      <a
+        onClick={onClickLink}
+        rel="noopener noreferrer"
+        key={result.id.raw}
```
joshua-scott, line 11: I guess this `key` can be removed if there are no complaints from React?

next/public/locales/en/common.json
```diff
@@ -40,6 +41,10 @@
   "statusmessage-success": "Success",
   "statusmessage-warning": "Warning",
   "statusmessage-error": "Error",
+  "language-switcher-button": "Language switcher button",
```
joshua-scott, line 44: If it's a button, screenreaders will announce it as such, so we can remove the text "button" from the translation itself

joshua-scott (commented):
This is looking really great! Lighthouse approves too :)

I left some comments with some small changes that may or may not make sense - let me know what you think.

CleanShot 2023-05-08 at 17 39 57

next/components/search/search-result.tsx
Zionknight-crypto: Thanks for pointing this out! Actually as far as I know React wants there to be unique keys for list items especially when generated dynamically to keep track of them so I moved that to the <li> element.

next/public/locales/en/common.json
Zionknight-crypto: You were right here, made the change.

next/components/main-menu/main-menu.components.tsx
Zionknight-crypto: This is pure gold! Cheers for this!

next/components/language-switcher.tsx
Zionknight-crypto: Actually I tested this part now 4 times and for some reason the first visible language option in the span is never read out loud by the voice over software so I need to pass that to the aria-label at least to read the current language in the button.

next/components/article-teaser.tsx
Zionknight-crypto: I restructured this so that the link would be the wrapping element and therefore no need for aria labelling.

next/components/header.tsx
Zionknight-crypto: This has been removed now.

joshua-scott (approved):
This is a huge improvement for a11y 🥇

PR #85, opened 2023-05-05
Disable importing reference data.

tharna:
*Link to ticket: JIRA-123*

Don't import reference data on first deployment as we are anyway going to do a fresh install.
3. Validate that xyz is working.


vermario (approved):
no more import. Thanks! merging!

PR #86, opened 2023-05-05
NEX-40: Add a "Latest articles listing" paragraph

vermario:
This branch adds a paragraph that the content editors can place on a Page or Frontpage content type. The listing will display a list of articles on the frontend:

image

The content editors can assign a translatable title to the component, and set a limit for the listing (for example, 3 items)

The frontend component uses react query to display the listing. This has to happen in the client side because we cannot anticipate if the page will use or not the component, and server side rendering has to be done in `getStaticProps` or `getServersideProps`

The listing calls - using react-query - an API route which in turn will ask drupal for the content using next-drupal's methods.

The api response will return cache headers.

The api accepts a parameter to specify the limit of items to be returned.

The new paragraph has been added to the migration, so the usual reset is needed to test this PR.

Silta ENVs:
https://feature-nex-40-listing-next.next4drupal-project.dev.wdr.io
https://feature-nex-40-listing.next4drupal-project.dev.wdr.io


next/package.json
```diff
@@ -57,6 +59,7 @@
     "@typescript-eslint/eslint-plugin": "^5.58.0",
     "@typescript-eslint/parser": "^5.58.0",
     "autoprefixer": "^10.4.14",
+    "cypress": "^12.11.0",
```
vermario, line 62: This should not be here

next/pages/api/articles-listing.ts
```diff
@@ -0,0 +1,42 @@
+import { NextApiRequest, NextApiResponse } from "next";
+import { DrupalNode } from "next-drupal";
+import { drupal } from "lib/drupal";
+
+import { validateAndCleanupArticleTeaser } from "@/lib/zod/article-teaser";
+
+import siteConfig from "@/site.config";
```
vermario, line 7: This is a bit strange, but it seems to work. We need to get to the default locale, and it's configured in that file.

next/pages/_app.tsx
```diff
@@ -15,21 +21,27 @@ import {
...
-          <Layout menus={menus}>
-            <Component {...restPageProps} />
-          </Layout>
-        </LanguageLinksProvider>
-      </Fonts>
+      <QueryClientProvider client={queryClient}>
+        <Fonts>
+          <LanguageLinksProvider languageLinks={languageLinks}>
+            <Layout menus={menus}>
+              <Hydrate state={pageProps.dehydratedState}>
```
vermario, line 38: I am not 100% sure if this is really needed if we use only client side queries?

next/components/articles-listing.tsx
```diff
@@ -0,0 +1,47 @@
...
+  listingId,
+  limit,
+}: {
+  listingId: string;
+  limit: number;
+}) {
+  const { t } = useTranslation();
+  const router = useRouter();
+  const { data, isLoading } = useQuery(
+    [`articles-${router.locale}-${listingId}`],
```
vermario, line 19: This is needed so we have a unique query for paragraph and language

Zionknight-crypto (approved):
I pulled the branch and ran it locally, code looks great, no errors in local or feature env. The only thing I could not check was the backend and changing content because I still can't login into the backend every time. None of my passwords work and if I have a failed login once it is always impossible to login in even with drush uli.

next/pages/api/articles-listing.ts
joshua-scott: Yeah, I think it works. We import it also in `language-links-context.tsx` and haven't seen any issues with it

next/pages/_app.tsx
joshua-scott: I'm not sure either. I suppose if we don't notice any difference then we can just remove it, since adding it there implies it's needed. But it depends what role react-query will eventually play in this project, maybe we will need to add it soon enough anyway?

joshua-scott (approved):
Everything is working great for me locally. Left a couple of comments with some food for thought, but I wouldn't be against merging this as-is.

next/pages/_app.tsx
vermario: Looks like we don't need the Hydrate part, I have removed it.

PR #88, opened 2023-05-05
feature/nex-11-adminpwd

vermario:
Changes proposed in this PR: use a circleCI environment variable to set tjthe admin password for drupal with `drush si` to avoid declaring the password to the world.

At this link you can see that the password is not output anymore after installation.
https://app.circleci.com/pipelines/github/wunderio/next4drupal-project/317/workflows/978fdde6-6305-4e5c-98d1-7d5b86a5c40c/jobs/1281?invite=true#step-125-88


tharna (approved):
Great!

PR #92, opened 2023-05-10
NEX-66: Add storybook build to silta deployments

vermario:


tharna (approved):
Looks good. Great job!

tharna:
/app/web needs to be there, that's where the storybook is built into.

PR #97, opened 2023-05-15
Nex-68: improved search page accessibility

Zionknight-crypto:
[Ticket](https://wunder.atlassian.net/browse/NEX-68)

- Restructure search results facets, result heading levels and remove props giving console errors.

*How to test:*
1. Open the link, perform a search and check console.
2. Use voice over after to listen for any strange announcements.
3. Use Wave a11y extension to check for errors (Ignore broken aria ref in search form - comes with the component)

[test link](https://feature-nex-68-searcha11y-next.next4drupal-project.dev.wdr.io/search)


next/components/search/search-multicheckbox-facet.tsx
```diff
@@ -52,18 +52,21 @@ export function MultiCheckboxFacet({
...
                   onClick={() => (checked ? onRemove(value) : onSelect(value))}
+                  aria-label={`example_facet_${label}${getFilterValueDisplay(
+                    option.value
+                  )}-title`}
                 />
-                <label
+                <span
                   className="ml-2 text-sm text-steelgray"
-                  htmlFor={`example_facet_${label}${getFilterValueDisplay(
+                  title={`Filter by ${getFilterValueDisplay(
```
vermario, line 61: We should translate this?

Zionknight-crypto: That is a brilliant question. I feel that we could decide on static label which will be translated for each local. For example: Filter by, tags etc

next/components/search/search-multicheckbox-facet.tsx
```diff
@@ -52,18 +52,21 @@ export function MultiCheckboxFacet({
...
+                  )} (${option.count.toLocaleString("en")})`}
                 >
                   {getFilterValueDisplay(option.value)}{" "}
                   <span className="text-steelgray">
                     ({option.count.toLocaleString("en")})
                   </span>
-                </label>
+                </span>
               </div>
             </li>
```
joshua-scott, line 71: I'm not sure about these aria-labels

For the one on line 55, voiceover is announcing it literally like this:

CleanShot 2023-05-17 at 15 03 13@2x

And then for the span after the checkbox, don't we want it to remain as a `label`? 

I don't think we need to overcomplicate this. The {label} just before the list of checkboxes clearly says what the filter is ("Tags", "Content type"), so I feel like we're good as long as 1. the label for each checkbox and checked/unchecked status is clearly announced, and 2. the label and checkbox are associated with each other such that clicking the label also toggles its checkbox.

next/components/user-menu.tsx
```diff
@@ -72,8 +72,8 @@ export function UserMenu() {
     <div>
       <span className="sr-only">{t("user-menu")}</span>
       <button type="button" className="hover:underline" onClick={toggle}>
-        <span className="hidden sm:mr-2 sm:inline">
-          {t("user-menu-account")}
+        <span className="sr-only capitalize sm:not-sr-only sm:mr-2 sm:inline">
+          {t("log-in-or-register")}
```
joshua-scott, line 76: This "log-in-or-register" string has been changed in main branch (it's just "account" now)

next/components/search/search-multicheckbox-facet.tsx
```diff
@@ -21,10 +21,10 @@ export function MultiCheckboxFacet({
 }: FacetViewProps) {
   const { t } = useTranslation();
   return (
-    <fieldset className={clsx(className, "mb-4")}>
-      <legend className="mb-5 text-heading-xs font-bold text-steelgray">
+    <section className={clsx(className, "mb-4")}>
```
joshua-scott, line 24: I think we can actually remove the `className` prop from this entirely - this component is only used in one place and the className isn't used

next/components/search/search-multicheckbox-facet.tsx
```diff
@@ -21,10 +21,10 @@ export function MultiCheckboxFacet({
...
+      <div className="mb-5 text-heading-xs font-bold text-steelgray">
         {label}
-      </legend>
+      </div>
       {showSearch && (
```
joshua-scott, line 28: I think we can also remove this showSearch prop and what it renders - it's not used at all afaik. (The search box is always visible at the top of the page, we'll never render it here with the filters)

next/components/search/search-multicheckbox-facet.tsx
```diff
@@ -43,28 +43,24 @@ export function MultiCheckboxFacet({
...
-                    ({option.count.toLocaleString("en")})
-                  </span>
-                </label>
-              </div>
+            <li
+              key={`${getFilterValueDisplay(option.value)}`}
+              className="flex items-center"
+            >
+              <Checkbox
+                data-transaction-name={`facet - ${label}`}
```
joshua-scott, line 51: Is this `data-transaction-name` useful here? I have a feeling it's also copied from the initial implementation and we can remove it too

next/components/search/search-multicheckbox-facet.tsx
joshua-scott: Also, I think a `fieldset` is better than a `section` here? Fieldset is used to group related form elements together, which is exactly what this is. I think `section` elements are for bigger building blocks of a page, not for this case where it's just a few form elements.

next/components/search/search-multicheckbox-facet.tsx
```diff
@@ -21,10 +21,10 @@ export function MultiCheckboxFacet({
 }: FacetViewProps) {
   const { t } = useTranslation();
   return (
-    <fieldset className={clsx(className, "mb-4")}>
-      <legend className="mb-5 text-heading-xs font-bold text-steelgray">
+    <section className={clsx(className, "mb-4")}>
+      <div className="mb-5 text-heading-xs font-bold text-steelgray">
         {label}
-      </legend>
+      </div>
```
joshua-scott, line 27: This `label` is actually pretty important - it's telling the user what the group of checkboxes is actually for, so I think we should make that clearer in some way. Is there a reason the fieldset+legend approach didn't work well? That's what MDN recommends for this kind of thing: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#try_it

If we want it to be clear that the search results are filtered by these values, we could also add to the aria-label a bit. Something like `aria-label={t("filter-by-label", { label })`, implemented in common.json like `"filter-by-label": "Filter by {{label}}"`

next/components/search/search-multicheckbox-facet.tsx
```diff
@@ -43,28 +43,24 @@ export function MultiCheckboxFacet({
...
+                aria-label={`${t("filter-by")} ${getFilterValueDisplay(
+                  option.value
+                )} (${option.count.toLocaleString("en")})`}
+                checked={checked}
+                onClick={() => (checked ? onRemove(value) : onSelect(value))}
+              />
+              <span className="ml-2 text-sm text-steelgray" aria-hidden={true}>
+                {getFilterValueDisplay(option.value)}{" "}
+                <span className="text-steelgray">
+                  ({option.count.toLocaleString("en")})
```
joshua-scott, line 61: Rather than using toLocaleString("en") with a hardcoded language, should we just drop it and use `count` directly? Or is there a reason we need it?

next/components/search/search-multicheckbox-facet.tsx
```diff
@@ -43,28 +43,24 @@ export function MultiCheckboxFacet({
...
-              </div>
+            <li
+              key={`${getFilterValueDisplay(option.value)}`}
+              className="flex items-center"
+            >
+              <Checkbox
+                data-transaction-name={`facet - ${label}`}
+                aria-label={`${t("filter-by")} ${getFilterValueDisplay(
+                  option.value
+                )} (${option.count.toLocaleString("en")})`}
```
joshua-scott, line 54: I think we can use `aria-labelledby` and point it to the text just after the checkbox, rather than duplicating it? (Of course, we'd remove the `aria-hidden` from it in that case)

The "filter by" part can be removed if we add "filter by" to the initial label as I suggested in an earlier comment - this would also prevent repeating "filter by" on every option, because the user would already know it's for filtering.

joshua-scott (changes requested):

PR #98, opened 2023-05-16
DO NOT MERGE: NEX-71: Demonstrate having a different language as default

vermario:
This draft PR is a test to show how to set a different language than english as default.

* We are setting finnish as selected language (not default) on the drupal side

* We are disabling automatic detection of next.js

* We are setting finnish as default language in next.js


vermario:
Closing this PR, we can have it as an example of how do achieve this.

PR #103, opened 2023-05-22
Nex 73: Logo section

Zionknight-crypto:
[*Nex-73*](https://wunder.atlassian.net/browse/NEX-73)

- Create a logo strip or section to display logos on a page.

*How to test:*
1. This is quite simple for now and using only svgs but later it can be adjusted to render other image files types as well.
2. You can check that the element is present and responsive.

[Link](https://feature-nex-73-logo-section-next.next4drupal-project.dev.wdr.io/)


next/components/logo-strip.tsx
```diff
@@ -0,0 +1,47 @@
...
+interface Logo {
+  image: ReactElement;
+  label: string;
+  id: string;
+}
+
+const logos: Logo[] = [
+  {
+    image: <Facebook className="inline-block h-16 w-16 text-primary-600" />,
+    label: "Share to Facebook",
```
joshua-scott, line 17: These labels would need translating

next/components/logo-strip.tsx
```diff
@@ -0,0 +1,47 @@
...
+    image: <LinkedIn className="inline-block h-16 w-16 text-primary-600" />,
+    label: "Share to LinkedIn",
+    id: "linkedIn",
+  },
+];
+
+export function LogoStrip() {
+  const { t } = useTranslation();
+  return (
+    <section id="logo-section">
```
joshua-scott, line 35: What's the id for here?

joshua-scott (commented):
This is a nice base for the upcoming project where we'll need to add this logo section. 

I think the choice of logos for this template can be considered - it's a bit confusing to have the social media logos that aren't used for sharing or links. 

One idea would be to add some of the same logos we use on wunder.io (just above the footer). That would also make a bit clearer how this should be used. Thoughts?

image

next/components/logo-strip.tsx
Zionknight-crypto: Good catch, I actually copied this code from the social share branch to begin with and now I reckon we won't need to translate these as it will say simply "CompanyX logo" for now.

next/components/logo-strip.tsx
Zionknight-crypto: Removed in latest commits.

joshua-scott (approved):

PR #105, opened 2023-05-23
Nex 75: Contacts list

Zionknight-crypto:
[*Nex 75*](https://wunder.atlassian.net/browse/NEX-75)

- Created a contact person list section

*How to test:*
1. It's tricky with the type of image you may use, however, any portrait with some space around the head of the person should work fine. Play with the screen sizes to see if there are any breaks as well. 

[Link](https://feature-nex-75-contacts-next.next4drupal-project.dev.wdr.io/)


next/components/contact-list.tsx
```diff
@@ -0,0 +1,75 @@
...
+              <NextImage
+                src={contact.image}
+                width={100}
+                height={100}
+                alt={`Portrait of ${contact.name}`}
+                className="circle-clip"
+              />
+            </div>
+            <p className="font-bold">{contact.name}</p>
+            <p>{contact.title}</p>
```
joshua-scott, line 60: Similar to `mailto:`, we can also use `tel:` here

next/components/contact-list.tsx
```diff
@@ -0,0 +1,75 @@
...
+    phoneNumber: "+358445123457",
+    email: "jane.dean@mail.com",
+    id: 2,
+  },
+];
+
+export function ContactList() {
+  const { t } = useTranslation();
+  return (
+    <section id="contacts-section" className="pb-8 pt-8">
```
joshua-scott, line 43: Padding can be simplified to `py-8`

next/components/contact-list.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <h2 className="text-heading-sm font-bold md:text-heading-md">
+        {t("contact")}
+      </h2>
+      <ul className="flex flex-wrap justify-around pb-4 pt-4">
+        {contacts?.map((contact) => (
+          <li key={contact.id} className="flex flex-col items-center p-4">
```
joshua-scott, line 49: Rather than using flex here, I think a grid might be better to ensure the orphaned child is left-aligned rather than in the center? (Also, I think the images could do with being slightly bigger on desktop, maybe?)

image

next/components/contact-list.tsx
```diff
@@ -0,0 +1,75 @@
...
+  id: number;
+}
+
+const contacts: Contact[] = [
+  {
+    image: "/john_dean.jpg",
+    name: "John Dean",
+    title: "CFO",
+    phoneNumber: "+358445123456",
+    email: "john.dean@mail.com",
```
joshua-scott, line 19: Looks like `mail.com` is a real thing, let's instead use example.com which is reserved for examples?

joshua-scott (commented):
This works pretty well; added some comments with small suggestions.

I'm also thinking, should we not add support for editing this in the backend? I assume most sites would want that, rather than a hardcoded one like this is now.

next/components/contact-list.tsx
Zionknight-crypto: done

next/components/contact-list.tsx
Zionknight-crypto: done

next/components/contact-list.tsx
Zionknight-crypto: done.

next/components/contact-list.tsx
Zionknight-crypto: done

next/components/contact-list.tsx
```diff
@@ -0,0 +1,82 @@
...
+];
+
+export function ContactList() {
+  const { t } = useTranslation();
+  return (
+    <section className="py-8">
+      <h2 className="text-heading-sm font-bold md:text-heading-md">
+        {t("contact")}
+      </h2>
+      <ul className="grid auto-rows-max grid-cols-1 justify-items-center gap-4 py-4 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4">
```
joshua-scott, line 47: `lg:grid-cols-4` with only 3 items means this is left-aligned on large screens, let's remove that class?

joshua-scott (approved):
Looks good, I made one minor comment but after that feel free to merge

next/components/contact-list.tsx
Zionknight-crypto: On it.

PR #106, opened 2023-05-24
Nex 76: Skip to content link and translations

Zionknight-crypto:
[*Next-76*](https://wunder.atlassian.net/browse/NEX-76)

- Create a visually hidden link which the user may navigate to using the voice-over software or tab button.
- On click, the user will be directed to main content.
- Currently, this does not look like much due to the type of content structure we have. However, it works simply.

*How to test:*
1. Test using tab button as well as voice-over software. You may press enter, click with the mouse or using the voice-over specified controls. 
2. Lighthouse accessibility showed no issues for this.

[Link](https://feature-nex-76-skip2content-next.next4drupal-project.dev.wdr.io/)


joshua-scott (changes requested):
Tested with keyboard and voiceover and it works well, but I think it can actually be much simpler:

- you can remove both refs that were created, neither of them are used as far as I can tell
- all the line changes in header.tsx are good. But I would move those lines to layout.tsx, because then the id="main-content" and <SkipToContentLink href="#main-content"> would be colocated (rather than an implicit connection between them, it’s simpler if they’re in the same file anyway). Less chance of it being changed and breaking functionality
- SkipToContentLink doesn’t seem to actually use onClick, onKeydown, handleClick or handleKeyDown at all - I would remove them unless they would actually be needed. Currently the conditionals such as `if (mainContentRef.current)` don’t do anything because the ref is not actually set there, so .current is always null

next/components/layout.tsx
```diff
@@ -17,7 +21,10 @@ export interface LayoutProps {
 }
 
 export function Layout({ menus, children }: LayoutProps) {
+  const mainContentRef = useRef<HTMLDivElement>(null);
```
joshua-scott, line 24: This mainContentRef is still not really used (I mean it's passed to the `main` element, but for no reason currently). Let's remove it!

next/wunder-component-library/skip-to-content-link.tsx
```diff
@@ -0,0 +1,19 @@
...
+interface ContentLinkProps {
+  href: string;
+  children: React.ReactNode;
+}
+
+export function SkipToContentLink({ href, children }: ContentLinkProps) {
+  return (
+    <Link
+      href={href}
+      className="sr-only font-inter tracking-wide text-primary-600 underline-offset-4 focus-visible:not-sr-only focus-visible:underline focus-visible:ring focus-visible:ring-primary-600"
```
joshua-scott, line 13: This works nicely but as discussed, we can probably move most of these focus-visible styles as they're handled globally by _focus-visilble.css. (It might be that `focus-visible:not-sr-only` is the only required one?)

joshua-scott (approved):
Added a couple of comments about things that can potentially be removed, but once you've had a look at that feel free to merge :)

PR #107, opened 2023-05-29
NEX-74: all articles page with pagination

vermario:
This branch adds: 

* an /all-articles page that shows all articles in a paginated view
* a link on the homepage taking the user to that page
* new functions to get articles in a paginated way
* some changes to the migration
   * added dates for the articles, otherwise they were all created on the same second and drupal was confused when listing them
   * added 6 more articles and translations, adjusted the tags (we have now drupal and next.js as tags)
   
   For the listing to work, I had to add the jsonapi_extras module so the response would contain a total count.
   
   
#### Locally
* Run the usual big command
* visit the frontpage and click on the all articles button
* see the list of articles, change page, confirm that the order is correct (you can use the date for this, the articles have dates in consecutive days
* create a new article, see that it appears in the list

## On silta

Drupal: https://feature-nex-74.next4drupal-project.dev.wdr.io
Next.js: https://feature-nex-74-next.next4drupal-project.dev.wdr.io

## Screenshots

image

image


drupal/composer.json
```diff
@@ -22,6 +22,7 @@
         "drupal/core-composer-scaffold": "^10.0",
         "drupal/core-recommended": "^10.0",
         "drupal/elasticsearch_helper": "^7.0",
+        "drupal/jsonapi_extras": "^3.23",
```
vermario, line 25: This is needed to get the count of items.

drupal/recipes/wunder_next_setup/config/next.next_entity_type_config.node.article.yml
```diff
@@ -10,4 +10,4 @@ configuration:
 revalidator: path
 revalidator_configuration:
   revalidate_page: true
-  additional_paths: "/\r\n/fi\r\n/en\r\n/sv"
+  additional_paths: "/\r\n/fi\r\n/en\r\n/sv\r\n/all-articles\r\n/fi/all-articles\r\n/sv/all-articles"
```
vermario, line 13: added the paths of the page to the list of invalidated paths

drupal/web/modules/custom/wunder_democontent/migrations/nodes_articles.yml
```diff
@@ -39,6 +39,13 @@ process:
   field_tags:
     plugin: entity_generate
     source: Tag
+  created:
```
vermario, line 42: without this, and the changes in the csv file, all articles had the same date of creation (since it's granular only at the second level), and so they were not listed in the correct order. (this took a while to figure out :-) )

next/components/article-list-item.tsx
```diff
@@ -0,0 +1,43 @@
+import Image from "next/image";
```
vermario, line 1: This is a new way of showing the article

next/components/latest-articles.tsx
```diff
@@ -21,7 +27,22 @@ export function LatestArticles({ articles }: LatestArticlesProps) {
...
-      {!articles?.length && <p className="py-4">{t("no-content-found")}</p>}
+      <div className="flex items-center justify-center">
+        {!articles?.length && <p className="py-4">{t("no-content-found")}</p>}
+        {articles?.length && (
+          <LinkButtonWithStyles
+            href="/all-articles"
+            className="text-base mr-4 mt-4 inline-flex px-5 py-3"
+            variant="primary"
+          >
+            {t("all-articles")}
```
vermario, line 38: adding a button to the existing "latest articles" to take the user to the all articles listing

next/lib/get-node-page-json-api-params.ts
```diff
@@ -61,8 +61,9 @@ export function getNodePageJsonApiParams(resourceType: ResourceType) {
       "status",
       "metatag",
       "field_excerpt",
+      "path",
     ]);
   }
 
-  return apiParams.getQueryObject();
+  return apiParams;
```
vermario, line 68: **This is a breaking change**

I had to change this so that it returns the apiParams themselves, so I could call the `getQueryObject()` method later on. So I had to change the way this function was called everywhere.

next/lib/zod/article.ts
```diff
@@ -14,6 +14,7 @@ export const ArticleBaseSchema = z.object({
   }),
   title: z.string(),
   field_image: ImageShape,
+  field_excerpt: z.string().optional(),
```
vermario, line 17: I added the excerpt to the zod type for the article, so I could use it in the new article-listing-item component.

next/pages/all-articles/[[...page]].tsx
```diff
@@ -0,0 +1,105 @@
...
+    fallback: "blocking",
+  };
+};
+
+export const getStaticProps: GetStaticProps<AllArticlesPageProps> = async (
+  context
+) => {
+  // Get the page parameter:
+  const page = context.params.page;
+  const currentPage = parseInt(Array.isArray(page) ? page[0] : page || "1");
```
Comment (vermario) [inline next/pages/all-articles/[[...page]].tsx:65]: this is a bit funky, but basically it you go to /all-articles without the "1" at the end this will turn the parameter into page 1

next/pages/index.tsx
```diff
@@ -64,6 +64,7 @@ export const getStaticProps: GetStaticProps<IndexPageProps> = async (
       "fields[node--article]": "title,path,field_image,uid,created",
       include: "field_image,uid",
       sort: "-created",
+      "page[limit]": 3,
```
vermario, line 67: fewer items in the frontpage, since we now have the all-articles listing

next/components/pagination.tsx
```diff
@@ -0,0 +1,126 @@
...
+    });
+    const focusable = focusRestoreRef?.current.querySelector(
+      'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])'
+    );
+    if (focusable) {
+      (focusable as HTMLElement).focus({ preventScroll: true });
+    }
+  };
+
+  const handlePrevClick = (e) => {
```
juhov, line 69: If you wan't to be extra nice for the user, these click handlers could check if user intended to open the link in new tab with for example cmd+click and then just let the click go to the `<Link>`

juhov: naturally you would need to also change from `<button>` to `<Link>`

next/components/pagination.tsx
```diff
@@ -0,0 +1,126 @@
...
+    prevPageHref,
+    nextPageHref,
+  } = paginationProps;
+  const { t } = useTranslation();
+
+  const router = useRouter();
+  const [isLoading, setIsLoading] = useState<"forward" | "back" | false>(false);
+  const numbers = [currentPage, totalPages].filter((n) => !isNaN(n));
+
+  const restoreScroll = () => {
```
juhov, line 56: ℹ️ Restore focus too for the keyboard users

joshua-scott (approved):
Reviewed the code and tested according to instructions and this is working very nicely indeed. Check comments for a couple of small things to think about, but otherwise this is good to go.

One more unrelated thing I noticed when adding a new article: if you don't provide an image, Zod complains about it because `field_image` isn't `.nullable()`. Should we a) make the image optional, or b) enforce the image in the CMS side? (Same question applies to our client project - for now I made it nullable in that project).

next/components/latest-articles.tsx
```diff
@@ -1,7 +1,13 @@
+import Link from "next/link";
 import { useTranslation } from "next-i18next";
 
 import { ArticleTeaser } from "@/components/article-teaser";
 import { ArticleTeaser as ArticleTeaserType } from "@/lib/zod/article-teaser";
+import ArrowIcon from "@/styles/icons/arrow-down.svg";
+
+import { withButtonStyles } from "@/wunder-component-library/button";
+
+export const LinkButtonWithStyles = withButtonStyles<any>(Link);
```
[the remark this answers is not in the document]
vermario: I made a ticket for it.

next/components/pagination.tsx
```diff
@@ -0,0 +1,126 @@
...
+      </MaybeLink>
+      {numbers.length > 0 && <p>{numbers.join("/")}</p>}
+      <MaybeLink href={nextPageHref}>
+        <Button
+          variant="tertiary"
+          aria-label="Seuraava"
+          disabled={!nextEnabled || !!isLoading}
+          onClick={handleNextClick}
+          tabIndex={props["aria-hidden"] ? -1 : undefined}
+        >
```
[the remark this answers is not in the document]
vermario: Fixed in 0f8a2deb1ecb65da24fdad202e4651b3797199bf (removed aria labels)

vermario:
> One more unrelated thing I noticed when adding a new article: if you don't provide an image, Zod complains about it because `field_image` isn't `.nullable()`. Should we a) make the image optional, or b) enforce the image in the CMS side? (Same question applies to our client project - for now I made it nullable in that project).

I have set them the field to nullable in 4290d40e86c804260c9c1646b263a0aa795f6a66

PR #108, opened 2023-05-31
NEX-77: Add language links for search page

vermario:
- add a new function that can be used on "next-only" pages, like the /search page, to calculate language links. Currently, switching language while on the search page takes the user to the frontpage. With these changes, the user is taken to the search page in the requested language.

*How to test:*
1. Go to the search page, try to switch languages.

### DIFF
--- a/next/pages/search.tsx
+++ b/next/pages/search.tsx
@@ -32,6 +32,10 @@ import { buildState } from "@/lib/search-ui-helpers/buildState";
 import { runRequest } from "@/lib/search-ui-helpers/runRequest";
 import { useNextRouting } from "@/lib/search-ui-helpers/useNextRouting";
 
+interface SearchPageProps extends CommonPageProps {
+  languageLinks: LanguageLinks;
+}
+
 export default function SearchPage() {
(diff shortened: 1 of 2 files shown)


next/pages/search.tsx
```diff
@@ -116,12 +124,17 @@ export default function SearchPage() {
   );
 }
 
-export const getStaticProps: GetStaticProps<CommonPageProps> = async (
+export const getStaticProps: GetStaticProps<SearchPageProps> = async (
   context
 ) => {
+  // We have to pass the path of the page manually here, it would be better to infer it
+  // automatically, but apparently this is not doable in this context.
+  const languageLinks = createLanguageLinksForNextOnlyPage("/search", context);
```
vermario, line 132: it's uncool that we have to hadcode the path here, since it comes from the filename anyway, but apparently there's no way to get to that information in this context.

vermario:
Maybe we will have to shelf this for a while, since this introduces another problem, where if you switch laguages after you have gotten some search results they stay in the previous language.

PR #110, opened 2023-06-01
NEX-77: Language links for next-only pages

vermario:
Currently, for next-only pages, when switching languages the user is taken to the homepage for that language: 

1. user is on `/all-articles`
2. user switches to Swedish using the language selector
3. user is taken to `/sv`

This PR changes this so that the user is taken to `sv/all-articles`

This needs to be applied manually to each page, and has these limitations: 
1. The path of the page needs to be hard-coded, because in the context of `getStaticProps` we can't know the path of the page (as weird as it feels, this seems to be the case)
2. We cannot add this for the `/search` page, because switching languages does not refresh the search results, and so the user would get an inconsistent UI. So for the search page, switching languages will take people to the frontpage.

So this PR adds this only for the `all-articles` page, but it can then be extended in the future to other pages that we will make.

### DIFF
--- a/next/pages/all-articles/[[...page]].tsx
+++ b/next/pages/all-articles/[[...page]].tsx
@@ -7,6 +7,10 @@ import { HeadingPage } from "@/components/heading--page";
 import { LayoutProps } from "@/components/layout";
 import { Meta } from "@/components/meta";
 import { Pagination, PaginationProps } from "@/components/pagination";
+import {
+  createLanguageLinksForNextOnlyPage,
+  LanguageLinks,
+} from "@/lib/contexts/language-links-context";
 import { getLatestArticlesItems } from "@/lib/get-articles";
 import { getCommonPageProps } from "@/lib/get-common-page-props";
 import {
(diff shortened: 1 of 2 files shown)


next/pages/all-articles/[[...page]].tsx
```diff
@@ -85,6 +90,11 @@ export const getStaticProps: GetStaticProps<AllArticlesPageProps> = async (
       : prevEnabled && [pageRoot, prevPage].join("/");
   const nextPageHref = nextEnabled && [pageRoot, nextPage].join("/");
 
+  // Create language links for this page.
+  // Note: the links will always point to the first page, because we cannot guarantee that
+  // the other pages will exist in all languages.
+  const languageLinks = createLanguageLinksForNextOnlyPage(pageRoot, context);
```
Comment (vermario) [inline next/pages/all-articles/[[...page]].tsx:96]: the `pageRoot` constant is already defined at line 84, and is used for the pagination links, so we can reuse it here.

(also note the comment, the language link will take users to the first page of the listing, which is a good feature I think).

joshua-scott (approved):
Tested this locally and it's working beautifully - let's merge it.

PR #111, opened 2023-06-01
NEX-80: File attachments paragraph

vermario:
This PR adds:

## Drupal:
1. a new paragraph called "File attachments" to which multiple media files can be attached. The paragraph can be added to frontpage, page, and inside accordions.
2. Migrations to add this component to the demo content (test pdf included)
(Due to a limitation of how we have the source files for the migration, only one file per paragraph is added... but the paragraph can accept any number of files and the frontend is aware of it)

## Next
1. New paragraph
2. Some changes about how we render the media items inside paragraphs (now using the generic <Media> component, which then renders the correct component based on the type of media passed.
3. A basic frontend listing of files, with a (fixed icon), file name - link , mime type and formatted size

### Screenshots

image

### To test

1. Visit the "Media and publishing" page, see the files displayed.
2. add some files, add a component to a page, see the files displayed

On Silta, this is the relevant page: 
(note that there is only one file as part of the migration (because reasons), but more can be added if you log into the drupal side and edit the content)

https://feature-nex-80-next.next4drupal-project.dev.wdr.io/en/media-and-publishing


next/components/media.tsx
```diff
@@ -1,16 +1,35 @@
...
 
 export function Media({ media, ...props }: MediaProps) {
   if (!media) {
     return null;
   }
 
+  function isFileAttachments(
+    media: MediaProps["media"]
+  ): media is FileAttachments["field_file_attachments"] {
+    return Array.isArray(media) && media[0]?.type === "media--document";
```
vermario, line 24: At the moment this is the only "media" that has multiple values. 

This function here will single out the file attachments, and render the appropriate component.

next/components/paragraph--image.tsx
```diff
@@ -1,6 +1,6 @@
-import { MediaImage } from "@/components/media--image";
+import { Media } from "@/components/media";
 import { Image } from "@/lib/zod/paragraph";
 
 export function ParagraphImage({ paragraph }: { paragraph: Image }) {
-  return <MediaImage media={paragraph.field_image} priority />;
+  return <Media media={paragraph.field_image} priority={true} />;
```
vermario, line 5: Since we have the generic <Media> component, I think we can use that instead of the more specific one in these components? 

(I have changed all of them)

next/components/media--file-attachments.tsx
```diff
@@ -0,0 +1,43 @@
...
+  ...props
+}: MediaFileAttachmentsProps) {
+  if (mediaItems.length === 0) {
+    return null;
+  }
+
+  return (
+    <>
+      <ul {...props} className="list-inside space-y-2">
+        {mediaItems.map((mediaItem) => (
```
vermario, line 21: I built this simple listing view with the info we have for the file. A more advanced version would be to have icons for different file types, and a fallback icon for when they are not recognized.

I wonder if we should just go ahead and add this more "advanced" version in the template?

next/components/paragraph--image.tsx
Zionknight-crypto: Seems good to me at this stage.

Zionknight-crypto (approved):
Looks great so far. Let's get it into the project and take it further.

Zionknight-crypto (approved):
Looks great so far. Let's get it into the project and take it further.

next/components/media--file-attachments.tsx
vermario: I made a new jira ticket for this, let's go with the current solution for now.

PR #115, opened 2023-06-16
Use supported mimetype for .mjs

MarttiR:
No ticket, 1-line fix.

- Use `text/javascript` instead of legacy `application/javascript`

*How to test:*
1. Check that Storybook still loads.

*Reasoning:*

https://datatracker.ietf.org/doc/rfc9239/

- Better align with current standards
    - `text/javascript` is now the only "proper" mimetype for JavaScript files
    - other types such as `application/javascript` alias to `text/javascript`
- Avoid magical thinking
    - If someone looking at silta-storybook.yml doesn't know about RFC9239, they might think that there has been some good reason to use specifically `application/javascript` for .mjs
    - This could cause them to start always setting that mimetype for .mjs or even all .js, without knowing why it's done or what is affected by doing so (= magical thinking)

### DIFF
--- a/silta/silta-storybook.yml
+++ b/silta/silta-storybook.yml
@@ -9,5 +9,8 @@ nginx:
   extraConfig: |
     include mime.types;
     types {
+        # Base Silta config uses "application/octet-stream" for .mjs.
+        # Browsers will refuse to run any JS files served with such mimetype.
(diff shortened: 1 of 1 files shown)


vermario (approved):
Thank you Martti! 

Storybook works as before, so merging this.

PR #117, opened 2023-06-20
NEX-86: more fields and config for accordion component

vermario:
This PR adds more fields and possibilities for the accordion paragraph.

1. there are now a text and link field for the accordion paragraph
2. there is also a layout field that can be used to select between one column and two columns layout
3. these fields have been added to the migration

Based on the layout field, two different react components are used.

#### Two columns:
image
https://feature-nex-86-next.next4drupal-project.dev.wdr.io/en/media-and-publishing

#### One column:

image

https://feature-nex-86-next.next4drupal-project.dev.wdr.io/en/media-and-publishing

#### One column with text and CTA:

image

https://feature-nex-86-next.next4drupal-project.dev.wdr.io/en/industries-using-drupal

#### Backend:

image


Zionknight-crypto (changes requested):
When I view the feature links I see that the wrapper container is not working for various paragraphs. Do you see this too?

vermario:
> When I view the feature links I see that the wrapper container is not working for various paragraphs. Do you see this too?

@Zionknight-crypto  try again. The reason is always the same: we have to wake up next _and_ drupal before reviewing. Both go to sleep after some hours for these release branches, and if we wake up only the next side, then it does not show images.

next/components/paragraph--accordion--two-columns.tsx
```diff
@@ -0,0 +1,70 @@
...
+            <div className="flex items-center justify-center py-3">
+              <LinkButtonWithStyles
+                href={paragraph.field_primary_link.full_url}
+                className="text-base mr-4 inline-flex max-w-sm px-5 py-3"
+                variant="primary"
+              >
+                {paragraph.field_primary_link.title}
+                <ArrowIcon
+                  aria-hidden
+                  className="bg:white ml-3 h-6 w-6 -rotate-90"
```
joshua-scott, line 46: This `bg:white` class isn't a thing (would be bg-white), let's remove it

joshua-scott (approved):
Looks good! I think it's a good decision to split this into two separate components - there's a bit of duplication but it's more readable that way.

I see you used `LinkButtonWithStyles`; at some point we should change this a bit so we'll just use the exported `buttonVariants` as a className on whatever component, rather than importing this higher-order component thing. But that can be done later - better to be consistent until we change that wherever else it's used.

next/components/paragraph--accordion--two-columns.tsx
Zionknight-crypto: I agree with Josh and I look forward to updating the variants/LInkButtonWithStyles soon once we have all current PRs merged.

Zionknight-crypto (approved):
Great work!

next/components/paragraph--accordion--two-columns.tsx
vermario: removed in : e0369eab4a9dd0238368cecebe4089ac7028b480 (it was in all the places where we use that arrowIcon component

PR #119, opened 2023-06-30
NEX-90: Add missing permissions to the next_api role to view all revisions

vermario:
* this fixes errors when viewing past revisions links.

There is one complication though, when viewing a past revision of the PAGE content type, that has paragraphs, the paragraphs are shown at their _current_ revision. (I think it's a limitation of jsonapi + paragraphs)

So in practice this is not really useful for content with paragraphs.

### Articles content type (no paragraphs, standard body field + image)

This works nicely!
image

### Page content type (made of paragraphs)

Paragraphs are shown at their last version in any case, making what is shown misleading for users:

image


vermario:
I have reported the issue with paragraphs to the next-drupal project here: https://github.com/chapter-three/next-drupal/issues/509

PR #122, opened 2023-07-04
NEX-90: Add patch to show paragraphs at correct revision when showing preview of a node revision

vermario:
With "vanilla" drupal core, when you request a node with paragraphs at a specified revision via JSONApi, the paragraphs will be returned at their most recent version. This is a problem for us here, because when a content editor clicks on a revision, they expect to see the node (and its paragraph) as they were at that point in history.

As suggested on https://github.com/chapter-three/next-drupal/issues/509 , we are now adding this patch:

* more info at: https://www.drupal.org/project/drupal/issues/3088239#comment-14708895

That seems to solve the problem nicely:

image

To test:

1. pick a page
2. edit the page, add an edit to a paragraph content (for example, add text like in the screenshot)
3. visit the "Revisions" tab
4. view the node at the original revision: you should not see the change that you have made in 2.
5. view the node at the current most recent revision, your change should be there.

### DIFF
--- a/drupal/composer.json
+++ b/drupal/composer.json
@@ -137,7 +137,8 @@
                 "Fix error when enabling the language module via recipe": "https://www.drupal.org/files/issues/2019-11-19/drupal.8.8.x-3002532-20.patch",
                 "Set MenuLinkContent getEntity to public visibility": "https://www.drupal.org/files/issues/2022-05-16/2997790-40.patch",
                 "Add link fields to jsonapi response": "https://www.drupal.org/files/issues/2022-03-04/3066751-68.patch",
-                "Do not set user as blocked when created via REST": "https://www.drupal.org/files/issues/2022-08-17/3055807-47.patch"
+                "Do not set user as blocked when created via REST": "https://www.drupal.org/files/issues/2022-08-17/3055807-47.patch",
+                "Return correct revisions of paragraphs in preview json api response": "https://www.drupal.org/files/issues/2022-09-27/3088239-33.patch"
             }
         },
         "dropin-paths": {


Zionknight-crypto:
I haven't been running N4D in a few weeks and tried the mega command after checking out to this branch. The preview mode doesn't seem to work, at one point there was my current revision and an old one from April if I recall and now the old one is gone and the current preview is not working. Moreover, the time stamps are incorrect, seems like they are 2-3 hours behind local time.

vermario:
> I haven't been running N4D in a few weeks and tried the mega command after checking out to this branch. The preview mode doesn't seem to work, at one point there was my current revision and an old one from April if I recall and now the old one is gone and the current preview is not working. Moreover, the time stamps are incorrect, seems like they are 2-3 hours behind local time.

I suggest testing this on the silta instance (your local always seems a bit strange :-) ) 

Drupal: 
`ssh www-admin@feature-nex-90-fix-shell.next4drupal-project -J www-admin@ssh.dev.wdr.io "drush uli"`

vermario:
I have tried this on the silta instance. After login, you should see:

List of revisions:
https://feature-nex-90-fix.next4drupal.dev.wdr.io/en/node/16/revisions

https://feature-nex-90-fix.next4drupal.dev.wdr.io/en/about-us <- This contains the "This is a test" text

https://feature-nex-90-fix.next4drupal.dev.wdr.io/en/node/16/revisions/16/view <- This, which is the older revision, does not thave the "This is a test" text

to me it looks like it's working as it should?

Tromppa:
I tested this now also on my local, and works fine! Only thing I'm missing is the Workflow or Content Moderation, where you can have a piece of content in eg. Draft mode before you publish that version, but I think this can be done later also. If you unpublish a revision now, the node becomes unpublished.

Tromppa (approved):

vermario:
> Only thing I'm missing is the Workflow or Content Moderation, where you can have a piece of content in eg. Draft mode before you publish that version, but I think this can be done later also. If you unpublish a revision now, the node becomes unpublished.


now that this works, we could look into enabling that. I wonder a bit whether it's a bit heavy to have it as part of the template or not though...

PR #126, opened 2023-07-09
NEX-92: Update core to 10.1.2 and contrib modules to latest available versions

vermario:
Main changes: 
* updated all contrib and core to 10.1.2
* removed patch for metatag, updated to 2.0
* added the restui module
* changed the passwords from plain text to hashed in users.csv because of https://www.drupal.org/node/3322420

To test: visit the site on silta: (wake up both if necessary)
* back https://feature-nex-92.next4drupal.dev.wdr.io/
* front https://feature-nex-92-next.next4drupal.dev.wdr.io

In my test the site works as before.


vermario:
Merging this.

PR #128, opened 2023-07-11
Updates. Rename UI components folder.

joshua-scott:
- Update all packages to the latest versions, but keep Next.js on version 13.3.x, because 13.4.x breaks search. (We should either investigate this further, or wait to see if this is fixed in a subsequent release)
- Rename "wunder-component-library" folder to simply "ui". "wunder-component-library" makes it sound more static and untouchable than it really is: it's _meant_ to be changed and added to etc.
- Since one of the updates is Prettier v3, which made `trailingComma: "all"` the default, this touches many files. But let's make that change now; I like the fact we just go with what prettier is doing and don't need to bikeshed about our prettier config.
- The updates also seems to fix an issue we were having with Storybook "cannot perform an update on an unmounted root" or similar. So that's nice.

To test, check the [feature environment](https://feature-updates-next.next4drupal.dev.wdr.io/) is working nicely, including login and search. If you feel like it, you _could_ checkout the branch locally and run `lando npm i && lando npm run lint && lando run build && lando npm run start` and `lando npm run storybook`.


Zionknight-crypto (approved):
Storybook looks great, ran locally and code seems just fine. Great update!

PR #134, opened 2023-08-15
NEX-94: Move the proxying for the sitemap from silta to next rewrites

vermario:
This Pr switches to next rewrites instead of silta redirects for the proxying of the sitemap.

This makes it simpler and more generic.


joshua-scott (approved):
Tested this and it's working great 👍

PR #30, opened 2022-12-23
NEX-31: Add partial support for redirects

vermario:
Support for redirects with multilingual setups seem to be a problematic area for this setup. I have written a message in the official slack: https://drupal.slack.com/archives/C01E36BMU72/p1671781046005709 

This PR adds partial support for redirects, with these limitations:

1. redirects need to be created manually, we cannot use the "Automatically create redirects when URL aliases are changed." `/admin/config/search/redirect/settings`, so that is disabled in the recipe
2. no redirects to external urls (https://www.drupal.org/project/decoupled_router/issues/3133681 old patch, won't apply)
3.  English redirects (or default language redirects) need to be marked for all languages
4.  language specific redirects need to be marked for that language

*Link to ticket: https://wunder.atlassian.net/browse/NEX-31*

*How to test:*

1. set the project up from scratch, following the instructions in https://github.com/wunderio/next4drupal-project/blob/feature/NEX-31/README.md 

### Redirects in default language

1. create an article with the title "test article" in English. It will get the path `/en/blog/test-article`
2. visit `https://next4drupal-project.lndo.site/en/admin/config/search/redirect/add` and add a redirect to the article like this: 
image

3. visit `https://frontend.lndo.site/en/redirected-path` or `https://frontend.lndo.site/redirected-path` and verify that you are redirected to `https://frontend.lndo.site/en/blog/test-article`

### Redirects in a non-default language

1. create an article with the title "test article finnish" in Finnish. It will get the path `/fi/blog/test-article-finnish` 
2. visit `https://next4drupal-project.lndo.site/fi/admin/config/search/redirect/add` and add a redirect to the article like this: 

image

3. visit `https://frontend.lndo.site/fi/redirected-path-finnish` and verify that you are redirected to `https://frontend.lndo.site/fi/blog/test-article-finnish`

PR #135, opened 2023-08-15
NEX-38: Support multilingual redirects, activate automatic creation of redirects when aliases are updated.

vermario:
This PR fixes the limitation that we saw in this previous PR around multilingual redirects: https://github.com/wunderio/next4drupal-project/pull/30

We were already using a patch for the decoupled router module, the latest version fixes the issue: https://www.drupal.org/project/decoupled_router/issues/3111456

Before this last version of the patch, if a redirect was found for a language, the calculated path returned to the frontend did not take into consideration the language requested. This patch fixes it, so we can reactivate the automatic creation of redirects when a path is updated. Since paths are tied to titles, this is very common.

## To test locally
1. rebuild the environment if testing locally using the big command
1. log into the backend
1. pick a content item, visit it on the frontend, take note of the path
1. change the title of the content, save. A new redirect should be automaticaly created
1. visit the old path in the frontend, you should be redirected automatically to the new path

## To test in silta

I have done the above, now this page: https://feature-nex-38-redirects-next.next4drupal.dev.wdr.io/fi/terveydenhuolto (that was the old url) should redirect to the new version https://feature-nex-38-redirects-next.next4drupal.dev.wdr.io/fi/terveydenhuolto-new

### DIFF
--- a/drupal/composer.json
+++ b/drupal/composer.json
@@ -126,7 +126,7 @@
                 "Get same results on different request": "https://www.drupal.org/files/issues/2019-07-18/change_request_type-63049395-09.patch"
             },
             "drupal/decoupled_router": {
-                "Unable to resolve path on node in other language than default": "https://www.drupal.org/files/issues/2022-12-01/decouple_router-3111456-resolve-language-issue-58--get-translation.patch"
+                "Unable to resolve path on node in other language than default": "https://www.drupal.org/files/issues/2023-04-27/decouple_router-3111456-resolve-language-issue-64--get-translation.patch"
             },
             "drupal/jsonapi_menu_items": {
                 "Add info about the langcode for menu items (issue #3192576)": "https://www.drupal.org/files/issues/2023-02-10/3192576-18.patch"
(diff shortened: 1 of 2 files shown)


drupal/composer.json
```diff
@@ -126,7 +126,7 @@
                 "Get same results on different request": "https://www.drupal.org/files/issues/2019-07-18/change_request_type-63049395-09.patch"
             },
             "drupal/decoupled_router": {
-                "Unable to resolve path on node in other language than default": "https://www.drupal.org/files/issues/2022-12-01/decouple_router-3111456-resolve-language-issue-58--get-translation.patch"
+                "Unable to resolve path on node in other language than default": "https://www.drupal.org/files/issues/2023-04-27/decouple_router-3111456-resolve-language-issue-64--get-translation.patch"
```
vermario, line 129: https://www.drupal.org/project/decoupled_router/issues/3111456#comment-15027425 This version of the patch contains the fix we needed.

drupal/recipes/wunder_next_setup/config/redirect.settings.yml
```diff
@@ -1,4 +1,4 @@
-auto_redirect: false
+auto_redirect: true
```
vermario, line 1: image

Zionknight-crypto (approved):
Just did the quick test with the silta link and this indeed does redirect.

PR #136, opened 2023-08-15
NEX-96: Change the request path for latest articles in each language

vermario:
* this means that the api call can be cached correctly when the site is behind a CDN

The frontend will now call a different url for each language: 
*  `/api/articles-listing/sv`
*  `/api/articles-listing/fi`
*  `/api/articles-listing/en`

So each of those calls will be cached differently.

I think it's useful to keep have caching on this call, otherwise Drupal can be exposed too much.


next/components/articles-listing.tsx
```diff
@@ -18,11 +18,14 @@ export function ArticlesListing({
...
     [`articles-${router.locale}-${listingId}`],
     async () => {
-      const response = await fetch(`/api/articles-listing?limit=${limit}`, {
-        headers: {
-          "accept-language": router.locale,
+      const response = await fetch(
+        `/api/articles-listing/${router.locale}?limit=${limit}`,
+        {
+          headers: {
+            "accept-language": router.locale,
```
vermario, line 25: the parameter that is actually used will still be taken from the header

Zionknight-crypto (approved):
Lets do it!

PR #139, opened 2023-08-23
NEX-98, NEX-99: Improvements and fixes for authentication in silta node

vermario:
We were using node 18.13 which seems to have issues with the fetch api (which has been introduced in node 18).

This caused all sorts of fun issues: 
1. no articles in listing in homepage
2. intermittent errors in authentication

The solution is to use a newer node version in the docker container which is used to run the frontend in silta.

During various attempts, I also updated NEXT_AUTH.

### DIFF
--- a/next/pages/api/auth/[...nextauth].ts
+++ b/next/pages/api/auth/[...nextauth].ts
@@ -47,7 +47,6 @@ export const authOptions: NextAuthOptions = {
   callbacks: {
     async jwt({ token, user, account }) {
       if (account && user) {
-        console.log(user);
         token.accessToken = user.access_token;
         token.accessTokenExpires = Date.now() + user.expires_in * 1000;
         token.refreshToken = user.refresh_token;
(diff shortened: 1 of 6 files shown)


next/pages/api/auth/[...nextauth].ts
```diff
@@ -24,17 +26,15 @@ export const authOptions: NextAuthOptions = {
...
         // Get access token from Drupal.
-        const response = await fetch(
-          `${env.NEXT_PUBLIC_DRUPAL_BASE_URL}/oauth/token`,
-          {
-            method: "POST",
-            body: formData,
-            headers: {
-              "Content-Type": "application/x-www-form-urlencoded",
-            },
+        const response = await drupal.fetch(url.toString(), {
```
Comment (vermario) [inline next/pages/api/auth/[...nextauth].ts:31]: seemed to make sense to use drupal.fetch everywhere

silta/node.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM wunderio/silta-node:18-alpine-v3.17
+FROM wunderio/silta-node:18-alpine-v0.1.2
```
vermario, line 1: This docker image has node 18.17

PR #147, opened 2023-10-02
Update to next.js 13.5.4,  update npm packages, test with lando 3.20.2

vermario:
Updates:

* next.js to 13.5.4
* various npm packages
* lando to 3.20.2 confirmed working

* this means that lando rebuild locally is needed *

Both Lando and NEXT promise better performances with these releases.

## To test
This is mostly about lando, so please pull, run the megacommand including the part that rebuilds lando.

Also, test specifically that the site works correctly with `lando npm dev` , the command does `lando npm build`


.lando.yml
```diff
@@ -139,7 +139,7 @@ services:
         DRUPAL_REVALIDATE_SECRET: revalidate_secret_not_secure_used_only_locally
         DRUPAL_CLIENT_SECRET: drupal_client_secret_not_secure_used_only_locally
         DRUPAL_CLIENT_ID: drupal-client-id
-        NEXT_PUBLIC_DRUPAL_BASE_URL: https://next-drupal-starterkit.lndo.site
```
vermario, line 142: using https with this version of `next dev` caused this issue:

```
 TypeError: fetch failed
    at Object.fetch (node:internal/deps/undici/undici:11457:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async exports.DrupalClient.fetch (/app/next/node_modules/next-drupal/dist/index.js:1:9978)
    at async exports.DrupalClient.getIndex (/app/next/node_modules/next-drupal/dist/index.js:1:17631)
    at async exports.DrupalClient.getEntryForResourceType (/app/next/node_modules/next-drupal/dist/index.js:1:18010)
    at async exports.DrupalClient.getResourceCollection (/app/next/node_modules/next-drupal/dist/index.js:1:15013)
    at async exports.DrupalClient.getResourceCollectionFromContext (/app/next/node_modules/next-drupal/dist/index.js:1:15458)
    at async getStaticProps (webpack-internal:///./pages/index.tsx:106:24) {
  cause: Error: unable to verify the first certificate
      at TLSSocket.onConnectSecure (node:_tls_wrap:1540:34)
      at TLSSocket.emit (node:events:513:28)
      at TLSSocket._finishInit (node:_tls_wrap:959:8)
      at ssl.onhandshakedone (node:_tls_wrap:743:12) {
    code: 'UNABLE_TO_VERIFY_LEAF_SIGNATURE'
  }
}
``` 

So, since this is only the local environment, we are switching to use the http version that lando also provides.

vermario:
https://github.com/vercel/next.js/releases/tag/v13.5.4 this is now out, we should test with this version, see if the issues with https are resolved.

.lando.yml
badrange: In our .env.local we have this setting which may or may not help you here:
```
# Needed for images to work locally
NODE_TLS_REJECT_UNAUTHORIZED=0 

```

vermario: hi @badrange ! 

thanks, that's the other option that we are considering. It does output a warning each time which is not great though. :/ 

I think the issue here is that we are doing everything inside lando, and it gets a bit complicated because lando has its own way to deal with https, which sounds like  it interferes with the newer and faster version of npm run dev

We'll try to fix this in the next days.

badrange: Cheers! Maybe you can find a way to suppress that warning; since it is only for local development it doesn't really matter. Seems a bit of a hassle to set up proper certificate trusting of lando's generated certificates for nodejs. But if you can manage to keep everything http only then sure that's probably even better.

vermario: There's a new version of next.js (13.15.4) and with that one the issue with https seems to have been fixed.

next/package.json
```diff
@@ -63,7 +63,7 @@
     "@typescript-eslint/eslint-plugin": "^5.62.0",
     "@typescript-eslint/parser": "^5.62.0",
     "autoprefixer": "^10.4.14",
-    "cypress": "^12.17.3",
+    "cypress": "12.17.3",
```
vermario, line 66: Let's fix the cypress version, because we are tied to the available docker image that we are using in lando, and they have an image with node18 and 12.17.3, but the one with 12.17.4 uses node 20 and we are not ready for that.

.lando.yml
vermario: It looks like in the newer version the issue is indeed fixed.

vermario:
@joshua-scott I have reduced the image size of the placeholder images we download from picsum to 1000*500. We don't show any image bigger than that in the site, so let's not use bandwith and disk for nothing :-) 

This should also speed up the migration on slower connections. We could also consider to ship the images in the gitpackage if we see that there's still a reliability problem with picsum.

.lando.yml
badrange: Good to know!

DavisAre (approved):
For now I generally tested & clicked around frontend & backend and seemed to work fine. 

`lando npm run dev` runs fine. For now tested on older lando version, so it is compatible also at least with lando v3.18.0 as well. @vermario

PR #148, opened 2023-10-03
Update the silta definitions for silta1.0

vermario:
This PR updates the silta setup according to the instructions posted at https://github.com/wunderio/silta/releases/tag/2023-10-03

I have changed the nginx dockerfile definition only, since we are using specially defined images for all other dockerfils.


Tromppa (approved):
Looks good!

PR #149, opened 2023-10-03
Sync drupal token expiration time, and session validity in frontend

vermario:
Before, the token was set to expire each 300 seconds, and the session in frontend was using the default value in next-auth, which is 30 days. 

This meant that it was likely to have a situation where the user was logged into the frontend, but the site used an expired token when doing calls to the backend.

both are now set to 5 days.

I think we should still investigate how to improve how the session is handled (for example checking it periodically and logging the user out if the token is expired and cannot be renewed etc), but this should be an improvement already.


joshua-scott (approved):
Looks good - I also made https://wunder.atlassian.net/browse/NEX-118 for the other improvements mentioned

PR #151, opened 2023-10-05
Use silta v1 major tags for docker images

joshua-scott:
This means the Docker images would be updated with the latest fixes. Related to Jānis' comment in slack [here](https://wunder.slack.com/archives/C8UN6AG9W/p1696507963658529?thread_ts=1696500393.268959&cid=C8UN6AG9W)


vermario (approved):
Looks good!

PR #152, opened 2023-10-05
NEX-119: Improvements for the big command to setup the template

vermario:
Currently the big command leaves the background node server running, and this can be confusing for first time users, that will need to run `lando npm-stop` to stop it, and are left without instructions.

In this PR we improve the command by: 

* stopping the background server operation after content has been imported into drupal via the migration
* provide better information to the terminal
* starting the server in production mode, giving also the alternative link to the frontend. The developer can then stop the server with control-c as usual, and nothing is running in the background
* small tweaks to the readme file

The terminal will show this output at the end of the big command: 

image


joshua-scott (approved):
Tested and everything is working perfectly - these changes make it much easier to follow what's happening too 👍

PR #153, opened 2023-10-06
Add script that runs all the commands required to start the project.

tharna:
## Link to ticket:

No ticket as I was just doing some ad hoc scripting

No feature environment needed, this is local development setup  command

This PR adds a script named start.sh which is able to run all the commends from the "mega command" and is also keeping track of the status so if any step fails the consecutive runs will continue from the last failed command instead of running all commands from the beginning.

## How to test:

1. Run ./start.sh in project root
2. Interrupt some command execution along the way (e.g. with CTRL-C)
3. Rerun the command and verify it continues from the step you interrupted it
4. Let it run to the end and verify you have fully working environment setup

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+
+# List of commands to run
+commands=(
+  "lando rebuild -y"
+  "lando composer install"
+  "lando generate-oauth-keys"
+  "lando drush si minimal -y"
+  "lando install-recipe wunder_next_setup"
+  "lando drush wunder_next:setup-user-and-consumer"
+  "lando drush eshd -y"
+  "lando drush eshs"
+  "lando npm i"
+  "lando npm run build"
+  "(lando npm run start&)"
+  "lando drush en wunder_democontent -y"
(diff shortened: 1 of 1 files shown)


vermario:
Really nice improvement, and it makes a lot of sense. Two things:

*  how about calling the script `setup.sh` instead of `start.sh`? It's more of a one-time thing rather than something you run every time you start working with this to my my mind
* in the meantime we have merged https://github.com/wunderio/next-drupal-starterkit/pull/152, where we did some changes, mostly to leave the user in a workable situation at the end of the script, and to tell them more what's happening. I can add those commands to the new script, if that's ok?

tharna:
Yes, setup.sh might make more sense :+1: 
Please go ahead and update the commands as needed to have them up to date.

vermario:
@tharna I have tweaked the script a bit, adding some more output, adding the starting of the frontend at the end of the script.

Here's how it looks like now: https://asciinema.org/a/612320

vermario (approved):
I think this is a great little improvement!

PR #159, opened 2023-11-10
Add tailwind intellisense VSCode extension

joshua-scott:
https://marketplace.visualstudio.com/items?itemName=bradlc.vscode-tailwindcss

This makes it much easier to use tailwindcss :)

### DIFF
--- a/next/.vscode/settings.json
+++ b/next/.vscode/settings.json
@@ -4,5 +4,9 @@
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true,
     "source.fixAll.stylelint": true
+  },
+  // Tailwind intellisense also in strings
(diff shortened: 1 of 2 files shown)


next/.vscode/settings.json
```diff
@@ -4,5 +4,9 @@
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true,
     "source.fixAll.stylelint": true
+  },
+  // Tailwind intellisense also in strings
```
vermario, line 8: comments are not allowed in json right? @joshua-scott

vermario: ok, VSCODE does not mind, and this file is only for VSCODE, so let's merge it

vermario (approved):

PR #161, opened 2023-11-20
Add .gitattributes file; update .gitignore

MarttiR:

Adds `.gitattributes` file, which tells git explicitly how to handle line ending normalisation for various file types, and by doing that, improves cross-platform developer experience.

## How to test:

1. Have a Windows developer make changes to some file.
2. See that there is no git warning about line endings when committing.
3. Also see that the line endings in the repository are indeed normalized to LF-only (Unix format `\n` instead of the Windows format `\r\n`)


vermario (commented):
I installed git on a "fresh" windows 11 machine, and tried changing a file (in vscode) and committing it. I did not get a warning either in this branch or on the main branch.

In the installation process (I got the installer from the official git page), there was this option selected as default: 

image

Which probably prevents this kind of issues. But, I think these changes would override the local settings for this repository in particular even if the developer had selected some other value. I think we can merge this.

vermario (approved):

PR #167, opened 2023-11-23
NEX-118: Improve handling of expired login sessions

joshua-scott:
## Link to ticket:

https://wunder.atlassian.net/browse/FAIS-44

Previously, if a logged in user's session expired, it was problematic because the frontend didn't "react" to it properly. If a logged-in user's session expired and they tried to access a protected route, they would be kicked to the frontpage (always in English), but the UI in the header would still show their username and give no indication that they aren't actually logged in anymore. The only way to resolve it was for the user to guess that logging out and back in would solve the problem.

This PR improves the situation like so:

If a logged-in user with an expired access token tries to visit `/dashboard`, which is a protected route, they are instead redirected to the `/auth/login` page in the correct language. Additionally, a new query param `logout=true` is added, which the login page uses to log the user out after the page has loaded on the client. This means it's very clear for the user that they have been logged out and must login again.

Note, rather than logout using the "logout" query param, it would be nice if we could simply log the user out in the server-side instead somehow. However, it seems [that's not possible](https://github.com/nextauthjs/next-auth/discussions/5334), so we log the user out clientside using the query param instead.

## Changes NOT proposed in this PR:

It would be nice to also ensure the sessions are refreshed at appropriate intervals using the refresh token, to minimise the chance of this happening in the first place. This PR doesn't address that, however - I'll look into that next and leave the ticket open in Jira for now.

## How to test:

1. Watch the helpful video I just sent you on slack!
2. To test locally, it's useful to change the maxAge to e.g. `10` (seconds) [here](https://github.com/wunderio/next-drupal-starterkit/blob/main/next/pages/api/auth/%5B...nextauth%5D.ts#L17). That way you can login and see what happens without waiting too long for the token to expire.
3. Be sure to test different languages too to ensure the redirects work as expected.

### DIFF
--- a/next/components/header/user-menu.tsx
+++ b/next/components/header/user-menu.tsx
@@ -72,7 +72,7 @@ export function UserMenu() {
       <span className="sr-only">{t("user-menu")}</span>
       <button type="button" className="hover:underline" onClick={toggle}>
         <span className="sr-only capitalize sm:not-sr-only sm:mr-2 sm:inline">
-          {t("account")}
+          {t("user-menu-account")}
         </span>
         <AccountIcon className="inline-block h-6 w-6" />
       </button>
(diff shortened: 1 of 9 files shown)


vermario (approved):
I tried this and it is a good improvement!

PR #170, opened 2023-12-13
NEX-128: Add limits for images in pixel size and megabyte size 

vermario:
Too big images can crash the site, this branch limits to: 
2500 x 2500 
and  5 megabytes. 

Larger images will be resized.

This applies to: 

- image in article
- image in media item
- images added to wysiwyg fields.


Tromppa (approved):
A very good and important improvement! :) I approve!

PR #171, opened 2023-12-15
Improve paragraphs filtering performance

MarttiR:

Improve paragraphs filtering performance by changing
```js
.filter(Boolean)
```
to simple null check function
```js
.filter((paragraph: any): boolean => paragraph != null)
```

Benchmarked to be around ~39% faster on 2-level nested paragraphs data, likely more with deeper nesting.

## How to test:

1. Check that front page and page content still load properly


vermario (approved):
nice! site still works :)

PR #174, opened 2024-01-19
NEX-134: Fix problem with deployments and long branch names

vermario:
## Link to ticket:

https://wunder.atlassian.net/browse/NEX-134

## What this adds
Fix the logic generating the special NEXT_DOMAIN and DRUPAL_DOMAIN env var values. This removes the mismatch between expected and actual domain values in case of a long branch names (which happens quite often) as those get truncated due to length limitation and a hash suffix is appended.

This solution was found by @k4lv15 for another project.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,6 +126,14 @@ commands:
             next_release_name=`silta ci release name --release-suffix next`
 
             # If name is too long, truncate it and append a hash
+            if [ ${#drupal_release_name} -ge 34 ]; then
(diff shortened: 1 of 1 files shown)


vermario:
are we happy with this @MarttiR @tharna ?

tharna:
Looks good :+1:  Classic off-by-one error :joy: 
This should also be backported to other templates / projects that might have this issue.

tharna (approved):

PR #180, opened 2024-02-20
NEX-121: update to node 20 in lando, circle ci and silta

vermario:
This PR includes these changes: 
### In lando 
 **minimum version of lando required [is now 3.21](https://github.com/lando/lando/releases/tag/v3.21.0-beta.1)**, because that's how we can use the node:20 service.  
 
 **Make sure you are using the latest lando. At the time of writing, lando3.21beta4 works** (There were some issues with memory limit in previous beta versions).
 
 Since there is a node20 version of the cypress image we were using previously, I have updated to that.
 
### In Circle ci
 We are switching to the new cicleci image that uses node20
 
###  In silta
We are switching to the alpine node 20 image in silta

## Update of next.js

Update next.js to latest 14.x version.

1. updated packages
2. updated configuration for images that generated an error 

## Update cypress

1. updated image, updated cypress version

> note: when running the tests locally,  I am seeing this error: 
``[1165:0220/120127.579341:ERROR:object_proxy.cc(590)] Failed to call method: org.freedesktop.portal.Settings.Read: object_path= /org/freedesktop/portal/desktop: unknown error type: ````

Seems like an issue with chromium, from quick googling. But the tests are run correctly anyway, so I would say we can ignore it.

## To test

1. update to latest lando 
2. run `./setup.sh`
(optional) 3. run cypress tests with `lando npm run cypress:run`

## Link to ticket:

https://wunder.atlassian.net/browse/NEX-121

https://nex-121-next.next-drupal-starterkit.dev.wdr.io/


next/components/main-menu/main-menu.tsx
```diff
@@ -94,15 +92,14 @@ export function MainMenu({ menu, isOpen, setIsOpen }: MainMenuProps) {
...
-            }
-          }),
+      menu.items?.some((item) =>
+        item.children?.some((subItem) => {
+          if (isMenuItemActive(router, subItem.url)) {
+            setActiveMenu(item.id);
+            setActiveSubmenu(null);
+            return true;
+          }
+        }),
```
vermario, line 102: Suddenly prettier was not happy with the previous formatting...

next/next.config.js
```diff
@@ -4,7 +4,13 @@ const { i18n } = require("./next-i18next.config");
...
   reactStrictMode: true,
   images: {
-    domains: [process.env.NEXT_IMAGE_DOMAIN],
+    remotePatterns: [
+      {
+        protocol: "https",
+        hostname: process.env.NEXT_IMAGE_DOMAIN,
+        pathname: "**",
+      },
+    ],
```
vermario, line 13: The domains setting was deprecated, and generated a warning when running the site.

@joshua-scott this happens in your project as well I think, maybe you want to do this same change?

.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ orbs:
 executors:
   cicd82:
     docker:
-      - image: wunderio/silta-cicd:circleci-php8.2-node18-composer2-v1
+      - image: wunderio/silta-cicd:circleci-php8.2-node20-composer2-v1.2
```
joshua-scott, line 9: Isn't it so that we should avoid using `v1.x` tag and just use `v1` unless there's a need for a specific point release? 

With the v1.2, my understanding is that it would always use image v1.2.x but not receive updates for v1.3, v1.4 etc ?

next/package.json
```diff
@@ -29,6 +29,7 @@
     "@t3-oss/env-nextjs": "^0.7.1",
     "@tanstack/react-query": "^5.8.4",
     "@tanstack/react-query-devtools": "^5.8.4",
+    "@types/react-dom": "^18.2.19",
```
joshua-scott, line 32: Maybe let's add `@types/react-dom` as a devDependency like the other type deps

joshua-scott (commented):
Looking good! Couple of comments. Don't have time to run it locally atm but should be able to do so tomorrow

next/package.json
vermario: fixed in 7a20b97c49968db1207b08e985bf64a2aa955f24

tormi:
My `./setup.sh` feedback:

```sh
 ✔ Container nextdrupalstarterkit_appserver_1  Started                                                                                          0.8s 
All settings correct for using Composer
Unable to write keys.dev.pub to: /var/www/.composer
ERROR ==> 
[+] Running 5/7
```

TBC

tormi:
2nd error during the setup:

```sh
[+] Healthchecking 1/1
 ✖ unknown shorthand flag: 'T' in -T
ERROR ==> unknown shorthand flag: 'T' in -T

[+] Running 1/1
 ✔ Container nextdrupalstarterkit_database_1  Started                                                                                           0.0s 
[+] Killing 1/1
 ✔ Container nextdrupalstarterkit_database_1  Killed 
```

vermario:
@tormi  I googled a couple of the errors, could it be that docker has run out of space? https://github.com/lando/lando/issues/2806

I have not seen these before :/

tormi:
Thanks, `docker system prune -a` gave _Total reclaimed space: 16.14GB_.

Now having that known memory issue:

```sh
➡️ Running command: lando install-recipe wunder_next_setup

Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 315392 bytes) in /app/drupal/web/core/lib/Drupal/Core/Database/StatementWrapperIterator.php on line 110
```

tormi:
Success! 
Screenshot 2024-02-22 at 17 45 30

phonoman:
After a good while of running into a wall because of the full disk space, it built after increasing it ✅ 
image

PR #184, opened 2024-02-22
NEX-121: Prevent lando memory limit errors during local development

joshua-scott:
(Note that PR base is NEX-121 branch, not main)

After adding this change, I no longer see errors like `Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 45056 bytes)` during drush commands. This error not happen with the previous version of lando but I've seen it frequently since the recent updates, in both this project and current client project.


vermario (approved):
I have not noticed this problem in my environment, but this is a good change.

PR #188, opened 2024-03-06
Add relalidate when returning a 404 for slug and frontpage

vermario:
When returning a 404 in pages that are generated using ISR, you should add a revalidate.

In this branch, we add 10 seconds for frontpage (since it's more critical) and the same 60 seconds for any page


Tromppa (approved):
Looks good to me!

PR #191, opened 2024-03-08
Improve service used to export the sitemap to a file

vermario:
In order to serve the sitemap at [frontend]/sitemap.xml , we: 

1. export the sitemap to a file in `public//siltemap.xml`
2. add a rewrite to next.js proxying /sitemap.xml to [drupal]/site/default/files/sitemap.xml

The current code in `wunder_sitemap` can result in the sitemap.xml file that gets exported to be deleted, since it's not registered as a drupal file.

With the changes in this PR:
* now the file will be an actual file entity with a usage, and so it won't be removed by cron.
* the file won't be replaced if the sitemap is empty

With this change, once you run cron, you should see this entry in the list of files:

image

### DIFF
--- a/silta/silta.yml
+++ b/silta/silta.yml
@@ -30,6 +30,7 @@ php:
       drush eshd -y
       drush eshs
       drush mim --group=demo_content --execute-dependencies --skip-progress-bar
+      drush cron
 
 # Configure reference data that will be used when creating new environments.
 referenceData:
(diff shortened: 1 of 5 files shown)


silta/silta.yml
```diff
@@ -30,6 +30,7 @@ php:
       drush eshd -y
       drush eshs
       drush mim --group=demo_content --execute-dependencies --skip-progress-bar
+      drush cron
```
vermario, line 33: This should make sure that the file exists when the starterkit is deployed in silta

Tromppa (approved):
Nice!

PR #192, opened 2024-03-14
NEX-143: Add pRetry to calls from frontend to Drupal

vermario:
## Link to ticket:

https://wunder.atlassian.net/browse/NEX-143

*  https://nex-143-next.next-drupal-starterkit.dev.wdr.io/
*  https://nex-143.next-drupal-starterkit.dev.wdr.io

Starting from a shared experience on the drupal next.js slack (https://drupal.slack.com/archives/C01E36BMU72/p1708638462166279?thread_ts=1708638293.534419&cid=C01E36BMU72) 

image

We are adding this npm package: 

https://www.npmjs.com/package/p-retry

And wrapping our calls to drupal with it, configuring it to retry 5 times in case of errors.

This should make the builds more resilient, expecially in the dev cluster in silta we have seen relatively often failed builds.

## How to test:

We should see fewer failures in builds (And during site operations), since the frontend should be able to retry to call Drupal if a request does not go well.


next/lib/drupal/get-menus.ts
```diff
@@ -1,18 +1,23 @@
...
+      Promise.all(
+        ["MAIN", "FOOTER"].map((menu) =>
+          drupal.doGraphQlRequest(GET_MENU, {
+            name: menu as MenuAvailable,
+            langcode: locale,
+          }),
+        ),
+      ),
+    { retries: 5 },
   );
```
joshua-scott, line 21: The thing with this one is, I think it'll retry if either of these `drupal.doGraphQlRequest`s fail?

If that's the case maybe it's better to structure it a bit differently so the pRetry deals with each promise individually and only retries whichever one failed? Something like this is what I'm thinking of (though I didn't try it):

```ts
const [main, footer] = await Promise.all(
  ["MAIN", "FOOTER"].map(menu =>
    pRetry(() => 
      drupal.doGraphQlRequest(GET_MENU, {
        name: menu,
        langcode: locale,
      }),
      { retries: 5 }
    )
  )
);
```

next/lib/drupal/get-articles.ts
```diff
@@ -22,12 +24,16 @@ export const getArticles = async ({
...
+    const articlesViewResult = await pRetry(
+      () =>
+        drupal.doGraphQlRequest(LISTING_ARTICLES, {
+          langcode: locale,
+          page: 0,
+          pageSize: limit,
+          offset: offset,
+        }),
+      { retries: 5 },
+    );
```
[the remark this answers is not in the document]
joshua-scott: Or maybe I nice "middle ground" for this is to create our own util function, `fetchWithRetry` or similar, that we can import and use everywhere it's needed? Advantage is less copy-pasting, and we wouldn't have to pass the `{ retries: 5 }` logic each time, and if we change that default it's only change in one place. Not a big deal either way but could be useful to do it like that maybe?

next/pages/dashboard/webforms/[webformName]/[webformSubmissionUuid].tsx
```diff
@@ -69,14 +70,18 @@ export const getServerSideProps: GetServerSideProps<
...
+      drupal.fetch(url.toString(), {
+        method: "GET",
+        headers: {
+          "Content-Type": "application/json",
+          // Pass the token to authenticate the request:
+          Authorization: `Bearer ${session.accessToken}`, // eslint-disable-line @typescript-eslint/no-base-to-string
+        },
+      }),
+    { retries: 5 },
+  );
```
Comment (joshua-scott) [inline next/pages/dashboard/webforms/[webformName]/[webformSubmissionUuid].tsx:84]: Testing locally, a "see more" submission link from my dashboard page (e.g. `https://frontend.lndo.site/dashboard/webforms/contact/86c2b13c-9aa6-4abe-bd00-197787fe46e1`) gives a 404 not found? Not sure if related to this new feature or not as the code LGTM here and the dashboard page itself works fine.

joshua-scott (commented):
I can definitely see the value in this!

Tested locally and works great, including when logged in and submitting a form, and viewing the dashboard. However see comments for a couple of questions/issues I had.

next/pages/dashboard/webforms/[webformName]/[webformSubmissionUuid].tsx
Comment (vermario) [reply next/pages/dashboard/webforms/[webformName]/[webformSubmissionUuid].tsx:84]: ah! it seems like it's an issue in the backend

I will investigate more, this is the matching issue in drupal.org: https://www.drupal.org/project/webform_rest/issues/3415211

Comment (vermario) [reply next/pages/dashboard/webforms/[webformName]/[webformSubmissionUuid].tsx:84]: ok, fixed by adding a patch in 5bf3e689c340dcfed8732ca9c5dc54d25f95ecc9

next/lib/drupal/get-menus.ts
vermario: ok! updated to your code (with a small addition) in 11999cfc066f1f2bef973a58705c997afe6399ec

next/lib/drupal/get-articles.ts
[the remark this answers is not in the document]
vermario: I had the same general idea (would be nice to specify those retries in one place only etc but I got lost trying to wrap the pRetry function one level up... :/

joshua-scott (approved):

PR #193, opened 2024-03-18
NEX-143: Retry functionality (revisited)

vermario:
in a previous PR ( #192 ), we added p-retry to allow the frontend to recover if drupal is not available momentarily.

But, we did it by wrapping each call to Drupal.fetch or drupal.doGrapqhlrequest in p-retry, which is a bit cumbersome and also means that future developers need to remember to add it everywhere.

In this revisited PR instead we do it at the drupal.client level (and in our added method to do graphql requests).

Also, we make sure that we don't use fetch by itself (we had a couple of API calls that did it) , but always drupal.fetch to get this added functionality.

Here's a short demo video: 

https://github.com/wunderio/next-drupal-starterkit/assets/185412/e3360054-582d-4923-853f-581fae98639f

### DIFF
--- a/next/lib/drupal/drupal-client.ts
+++ b/next/lib/drupal/drupal-client.ts
@@ -1,8 +1,31 @@
+import pRetry from "p-retry";
+
 import { GraphQlDrupalClient } from "./graphql-drupal-client";
 
 import { env } from "@/env";
 
+const getFetcher =
+  () =>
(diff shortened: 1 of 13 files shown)


joshua-scott (approved):
Didn't run locally this time but watched the (very helpful!) demo and the feature env is looking good, very nice improvement 👍

PR #195, opened 2024-03-22
NEX-144: Switch to standalone next build in CI, adapt the dockerfile

vermario:
In this PR: 

* set the  [standalone option](https://nextjs.org/docs/pages/api-reference/next-config-js/output#automatically-copying-traced-files) for `npm run build` when in CI
* adapt the dockerfile to copy the needed directories in the right place, and switch to the minimal `server.js` script to run the site 

This gives us: 
1. smaller docker image size
2. quicker build ( before: 1m 35s, after: 32s)

should be greener!

### To test: 

1. on silta the site should still work: https://standalone-next.next-drupal-starterkit.dev.wdr.io
2. locally, nothing should change, you can still use `npm run build` and then `npm run start`

### DIFF
--- a/silta/node.Dockerfile
+++ b/silta/node.Dockerfile
@@ -10,5 +10,5 @@ WORKDIR /app
 
 ENV PORT 3000
 
-# In standalone mode, we cannot use npm run start, we have to use the server.js file
+# server.js is created by next build from the standalone output
 CMD ["node", "server.js"]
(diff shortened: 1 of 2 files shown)


silta/node.Dockerfile
```diff
@@ -1,5 +1,14 @@
...
+COPY ./.next/standalone /app
+COPY ./.next/static /app/.next/static
+COPY ./public /app/public
 
-CMD npm run start
+WORKDIR /app
+
+ENV PORT 3000
+
+# In standalone mode, we cannot use npm run start, we have to use the server.js file
```
juhov, line 13:
```suggestion
# server.js is created by next build from the standalone output
```

vermario: Wait, I didn't know about this way of suggesting changes! (where have I been!) :D

PR #196, opened 2024-03-26
NEX-147: Do not revalidate in silta when building the template

vermario:
Now that we added #193 , our build in silta for the main environment started failing.

The reason is that we run the migration included in `wunder_democontent` after deployment, which in turn tries to call the frontend for revalidation. This won't work, because the frontend will either not exist or have a different secret

So this pr adds a mechanism to prevent revalidation while the migration runs.

This is done with an event subscriber set to a higher priority than what the `next` module provides, which will stop the propagation of the event if a state variable is set. We set that variable during the build in silta, then we unset it.

Also, I added a couple of other small things: 

1. I removed the proxy elasticsearch in lando (gave an error, does not really do much)
2. we were checking for the env var `CI` when deciding whether to build in standalone mode, but apparently inside lando it's set to 1 as well, so we are switching for the more specific `CIRCLECI` which we can be sure is there in circle ci (https://circleci.com/docs/variables/#built-in-environment-variables)

After deployment, these logs are visible: 

image

Also in circle ci we can see no more guzzle errors.

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -198,8 +198,6 @@ services:
 proxy:
   mailhog:
     - mail.lndo.site
-  elasticsearch:
-    - elasticsearch.lndo.site:9200
   kibana:
     - kibana.lndo.site:5601
   node:
--- a/next/next.config.js
+++ b/next/next.config.js
@@ -4,7 +4,7 @@ const { i18n } = require("./next-i18next.config");
 const nextConfig = {
   reactStrictMode: true,
   // Only generate standalone output in circle ci:
-  output: process.env.CI ? "standalone" : undefined,
+  output: process.env.CIRCLECI ? "standalone" : undefined,
   images: {
     remotePatterns: [
       {
--- a/setup.sh
+++ b/setup.sh
@@ -15,6 +15,7 @@ commands=(
   "(lando npm run start&)"
   "lando drush en wunder_democontent -y"
   "lando drush mim --group=demo_content --execute-dependencies"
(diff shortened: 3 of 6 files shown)


.lando.yml
```diff
@@ -198,8 +198,6 @@ services:
 proxy:
   mailhog:
     - mail.lndo.site
-  elasticsearch:
-    - elasticsearch.lndo.site:9200
```
vermario, line 202: this failed locally with the latest version of lando. Can't think of a good reason to have this since localhost:9200 does the same. I removed it.

next/next.config.js
```diff
@@ -4,7 +4,7 @@ const { i18n } = require("./next-i18next.config");
 const nextConfig = {
   reactStrictMode: true,
   // Only generate standalone output in circle ci:
-  output: process.env.CI ? "standalone" : undefined,
+  output: process.env.CIRCLECI ? "standalone" : undefined,
```
vermario, line 7: apparently in my node container the `CI` var was set to 1, this more specific one is preferable

setup.sh
```diff
@@ -15,6 +15,7 @@ commands=(
   "(lando npm run start&)"
   "lando drush en wunder_democontent -y"
   "lando drush mim --group=demo_content --execute-dependencies"
+  "lando drush pm-uninstall wunder_democontent migrate migrate_tools migrate_plus -y"
```
vermario, line 18: no reason to keep that module (and dependencies) installed after migration

PR #198, opened 2024-03-27
NEX-148: Update to next.js 14.1.4 and dependencies

vermario:
This PR updates next to the [14.1.4](https://github.com/vercel/next.js/releases/tag/v14.1.4) version.

Now the npm command should show 0 vulnerabilities.

This brings improvements for the sitemap functionality, that we hope to use to move the sitemap functionality to the frontend.


next/package.json
```diff
@@ -67,14 +67,14 @@
     "@tailwindcss/aspect-ratio": "^0.4.2",
     "@tailwindcss/typography": "^0.5.10",
     "@types/node": "^18.18.12",
```
joshua-scott, line 69: (Since we're using node 20, makes sense to update `@types/node` to v20 too)

joshua-scott (approved):
Can't test locally atm but LGTM and we're using this version of Next.js without issues in our client project based on this starterkit. Looking forward to the hinted sitemap improvements!

next/package.json
vermario: Good point! Updated in https://github.com/wunderio/next-drupal-starterkit/pull/198/commits/367e0ebf31327c60c37fd37c12e90e4cb3d73ec7

PR #200, opened 2024-04-16
NEX-149: Update to next.js 14.2.1

vermario:

* update to next.js 14.2.1
* tweaking of the `./setup.sh` command
* small lando tweak for kibana

## How to test:

1. check the site works on silta
2. run the `setup.sh` command locally


setup.sh
```diff
@@ -10,13 +10,14 @@ commands=(
...
-  "(lando npm run start&)"
   "lando drush en wunder_democontent -y"
+  "lando drush state:set wunder_democontent.disable_revalidation TRUE"
   "lando drush mim --group=demo_content --execute-dependencies"
   "lando drush pm-uninstall wunder_democontent migrate migrate_tools migrate_plus -y"
-  "lando npm-stop || true"
+  "lando drush state:del wunder_democontent.disable_revalidation"
+  "lando drush cron"
+  "lando npm i"
+  "lando npm run build"
```
vermario, line 20: We can use here the same thing we do in silta, where we disable temporarily the revalidation. This way we can do things in a more natural order, and we don't need to start the frontend in the background etc.

joshua-scott (approved):
Tested by running ./setup.sh locally with latest lando, works perfectly 👍

PR #201, opened 2024-04-20
Replace kibana with Elasticvue in Lando

vermario:
We currently include Kibana in our Lando setup as a tool to monitor and do queries to Elasticsearch. 

My personal preference for Elasticsearch UIs is [ElasticVue](https://elasticvue.com/) , which is simpler to use and more suitable - in my experience - for the kind of tasks we might want to do in lando, like checking if indexes exists, run simple searches, inspect what is indexed etc.

Apparently it is also available as a docker image, so this PR adds it as  a lando service.

After rebuild, it will be available at http://elasticvue.lndo.site , and it will connect automatically to the elasticsearch service.

Here's how it looks like:

### Listing indices:

image

### Listing documents:

image

### Showing one document:

image

You can check more features here: https://elasticvue.com/features


.lando.yml
```diff
@@ -147,7 +147,7 @@ services:
   # This service adds elasticsearch, used by (you guessed it) the search part of
   # the application.
   elasticsearch:
-    type: compose
+    type: lando
```
vermario, line 150: Apparently in new versions of lando the `compose` service is [now deprecated](https://docs.lando.dev/plugins/compose/) and the `lando service` should be used instead.

setup.sh
```diff
@@ -3,7 +3,6 @@
 # List of commands to run:
 commands=(
   "lando rebuild -y"
-  "lando composer install"
```
vermario, line 6: `lando rebuild` will trigger `composer install` already, so we can skip this step.

Zionknight-crypto (approved):
All for it!

PR #203, opened 2024-04-24
WIP: Evaluating Gql.tada instead of graphql-codegen

vermario:
WIP, Experiment


vermario:
For now, we think graphql-codegen is a better fit for our project. Closing.

PR #204, opened 2024-04-24
TypeScript: Enable noErrorTruncation in TS config

joshua-scott:
This improves DX by revealing more of the type within e.g. VSCode and other IDEs that use TS language server features. Very handy indeed. Demo gif from another project that uses the starterkit:

image


Zionknight-crypto (approved):
All for it!

vermario (approved):
let's do it! :)

PR #205, opened 2024-04-24
NEX-151: Use the new Translations GraphQL field to get translated versions of the node

vermario:
The graphl_compose module has[ recently added a new feature](https://www.drupal.org/project/graphql_compose/issues/3432965) to include a `translations` field on nodes that can be used to get information of the current node in other languages.

This allows us to remove the current slower way of finding translations, based on doing a query for each defined language to see if the node is translated in that language.

This should allow us to save _A LOT_ of requests when building pages.

In CircleCI the build time takes 10 seconds less (1m 38s -> 1m 28s) but the part where the requests are done starts after about 30 secs, so the gain is quite big I think!

This PR:

1. adds a new fragment
2. modifies the existing function
3. removes the function that we used to have to do multiple requests

### To test: 

1. run ./setup.sh
2. wait
3. Navigate to a page, or article (The frontpage is not included in this), check that you are able to navigate to other language versions using the language switcher
4. try to delete a translation for a node, to see if this is supported correctly (in my test it was): for that language now you should get a link to the frontpage in that language.


next/lib/graphql/fragments/nodes.ts
```diff
@@ -41,6 +41,9 @@ export const FRAGMENT_NODE_ARTICLE = graphql(`
         ...FragmentUser
       }
     }
+    translations {
+      ...FragmentNodeTranslation
+    }
```
vermario, line 46: We have to add this fragment to each node type individually, since the property is not on the NodeInterface (makes sense, only some node types could be set as translatable).

joshua-scott (approved):

PR #206, opened 2024-05-01
Support rendering nodes that are not set up to be translatable

vermario:
Currently the recent changes in #205 assume that all node types are set up for translations.

This might not be the case, and currently the site will break, expecting all content types to have a `translations` property. this PR fixes that, allowing the node not to have that property and still be rendered.

The language links will just be the standard ones.

### DIFF
--- a/next/lib/contexts/language-links-context.tsx
+++ b/next/lib/contexts/language-links-context.tsx
@@ -13,14 +13,20 @@ const LanguageLinksContext = createContext(siteConfig.locales);
  * From the site config and available node translations, create links to be used in the language switcher.
  */
 export function createLanguageLinks(
-  nodeTranslations?: FragmentNodeTranslationFragment[],
+  nodeTranslations: FragmentNodeTranslationFragment[],
 ): LanguageLinks {
-  const languageLinks = JSON.parse(JSON.stringify(siteConfig.locales));
+  const languageLinks = getStandardLanguageLinks();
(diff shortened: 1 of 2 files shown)


next/lib/contexts/language-links-context.tsx
```diff
@@ -13,15 +13,21 @@ const LanguageLinksContext = createContext(siteConfig.locales);
...
   });
   return languageLinks;
 }
 
+/**
+ * Get the standard language links from the site config.
+ */
+export const getStandardLanguageLinks = () =>
+  JSON.parse(JSON.stringify(siteConfig.locales));
+
```
vermario, line 30: Made this into a function since we use it in multiple places

next/pages/[...slug].tsx
```diff
@@ -189,7 +190,14 @@ export const getStaticProps: GetStaticProps<PageProps> = async (context) => {
...
   // Add information about possible other language versions of this node.
-  const languageLinks = createLanguageLinks(nodeEntity.translations);
+  let languageLinks;
+  // Not all node types necessarily have translations enabled,
+  // if so, only show the standard language links.
+  if ("translations" in nodeEntity) {
+    languageLinks = createLanguageLinks(nodeEntity.translations);
+  } else {
+    languageLinks = getStandardLanguageLinks();
+  }
```
Comment (vermario) [inline next/pages/[...slug].tsx:200]: use default links if the node is not translatable

vermario:
Merging this.

PR #210, opened 2024-05-07
NEX-152: Return permanent or temporary redirect based on the redirect status code set on the Drupal side

vermario:
## Link to ticket:

https://wunder.atlassian.net/browse/NEX-145

We have setup the [..slug].tsx page to handle redirects: if the route corresponds to a redirect we do a redirect in next.js

However, so far we were always returning a temporary redirect (which for next.js means `307` response status). In drupal you are able to set the response code (the most common being 301). 

So in this PR we map status codes in Drupal to response codes in next.js like this: 

`302`, `307` -> `permanent: false`
all others -> `permanent: true`

image

### DIFF
--- a/next/pages/[...slug].tsx
+++ b/next/pages/[...slug].tsx
@@ -114,7 +114,9 @@ export const getStaticProps: GetStaticProps<PageProps> = async (context) => {
     return {
       redirect: {
         destination: redirect.url,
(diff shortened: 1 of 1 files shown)


vermario:
The same change was done in a project and approved there, I will merge this

PR #213, opened 2024-05-15
NEX-153: Add separate consumer for previewing, prevent anonymous access to graphql

vermario:
## Link to ticket:

https://wunder.atlassian.net/browse/NEX-153

https://nex-153-next.next-drupal-starterkit.dev.wdr.io/

Currently, following the original setup of `next-drupal` we are using one consumer in Drupal, which is associated with a role with higher privileges (bypass access control, execute graphql queries)

When doing calls from next.js to Drupal, currently we could decide whether to use authentication (so "becoming" this admin like user) or to call drupal without authentication. When building pages for preview we were using authentication (because we need to preview unpublished nodes, revisions), but when building pages for "normal consumption" we were using anonymous requests.

This meant though that the graphql endpoint needed to be left open for anonymous users, which is definitely not great.

So in this PR: 

1. we add a second role (next-api-viewer) that is able to use graphql but has no special permissions in terms of accessing unpublished content)
2. we add a second consumer (with its own id and secret) connected to this new role. Also, we connect the `frontend-login`role to this second consumer, so that it gets only "normal user" privileges.
3. in the frontend, we create two instances of the drupalClient, one for previewing, and one for normal operations, and switch between the two when necessary.
4. we then remove access to the graphql endpoint for anonymous users.

## How to test:

Locally, run `./setup.sh` , check that: 

1. the site builds
2. you are able to preview unpuslished contents

NOTE: login does not work in silta at the moment for unrelated issues


drupal/recipes/wunder_next_setup/config/require_login.settings.yml
```diff
@@ -15,7 +15,7 @@ requirements:
   request_path:
     id: request_path
     negate: true
-    pages: "/subrequests/*\r\n/jsonapi\r\n/jsonapi/*\r\n/router/translate-path\r\n/next/preview-url\r\n/oauth/*\r\n/wunder_search/proxy\r\n/webform_rest/*\r\n/user/register\r\n/user/reset\r\n/user/reset/*\r\n/graphql"
+    pages: "/next/preview-url\r\n/oauth/*\r\n/wunder_search/proxy\r\n/webform_rest/*\r\n/user/register\r\n/user/reset\r\n/user/reset/*"
```
vermario, line 18: we can now protect the graphql route with require_login since we won't do anonymous requests to it

drupal/recipes/wunder_next_setup/config/user.role.next_api_role.yml
vermario, line 1: This is the role that existed before as well, I just added a couple of permissions, and added a label

drupal/recipes/wunder_next_setup/config/user.role.next_api_viewer_role.yml
vermario, line 1: This is our new "viewer" role

drupal/recipes/wunder_next_setup/recipe.yml
```diff
@@ -42,7 +42,6 @@ config:
         label: 'Anonymous user'
       grantPermissions:
         - 'restful post user_registration'
-        - 'execute graphql_compose_server arbitrary graphql requests'
```
vermario, line 45: no more access to graphql to the anonymous user 💯

drupal/web/modules/custom/wunder_next/src/Commands/WunderNextCommands.php
vermario, line 1: This is the drush command that we run during .setup.sh to set up the consumers and create the users

It used to allow options to be passed in etc, but we never really used it that way. I have simplified it to create the needed users and consumers without additional input

drupal/web/modules/custom/wunder_next/src/Commands/WunderNextCommands.php
```diff
@@ -15,115 +15,109 @@ class WunderNextCommands extends DrushCommands {
...
+      ],
+      // This second consumer is associated with a role with fewer permissions,
+      // and it is used when the frontend user does not need access to
+      // unpublished content.
+      [
+        'username' => self::API_VIEWER_USER_NAME,
+        'mail' => self::API_VIEWER_USER_MAIL,
+        'role' => self::API_VIEWER_USER_ROLE,
+        'secret' => $_ENV['DRUPAL_CLIENT_VIEWER_SECRET'],
+        'additional_roles' => [self::FRONTEND_LOGIN_ROLE],
```
vermario, line 57: this second consumer, for the "viewer" will also be used by the frontend login users, which now will not be able to see unpublished content

drupal/web/modules/custom/wunder_next/wunder_next.module
```diff
@@ -50,14 +50,19 @@ function wunder_next_consumer_load($entities) {
...
+
+    switch ($consumer->getClientId()) {
+      case $settings['client_id']:
+        $consumer->set('secret', $password_service->hash($settings['client_secret']));
+        break;
+
+      case $settings['client_viewer_id']:
+        $consumer->set('secret', $password_service->hash($settings['client_viewer_secret']));
+        break;
     }
```
vermario, line 66: we have added this code to be able to override the secrets at runtime, now that we have two consumers we need to do it for both of them

next/app/sitemap.ts
```diff
@@ -26,14 +26,10 @@ export default async function sitemap(): Promise<MetadataRoute.Sitemap> {
...
-        {
-          page: page,
-          langcode: lang,
-        },
-        false, // We only want published nodes, so we disable authentication.
-      );
+      const data = await drupal.doGraphQlRequest(GET_SITEMAP_NODES, {
+        page: page,
+        langcode: lang,
+      });
```
vermario, line 32: This is a good example, we used to disable authentication, now we can just leave the default which is to use the authentication. And because we are using the `drupal` client, the request will be done using the `next-api-viewer` role, which gets only published content

next/lib/drupal/drupal-client.ts
```diff
@@ -23,12 +23,31 @@ const getFetcher =
...
+  {
+    // We specify our own fetcher to apply pRetry to it:
+    fetcher: getFetcher(),
+    forceIframeSameSiteCookie: env.NODE_ENV === "development",
+    auth: {
+      clientId: env.DRUPAL_CLIENT_ID,
+      clientSecret: env.DRUPAL_CLIENT_SECRET,
+    },
+  },
+);
```
vermario, line 53: I think we have to do it this way (creating two instances) instead of passing a different auth each time like the docs would allow, because our doGrapqhlQuery method does not have access to that information since it's marked as private in the class...) 

This way seems to work :-)

next/pages/[...slug].tsx
```diff
@@ -99,13 +99,14 @@ export const getStaticProps: GetStaticProps<PageProps> = async (context) => {
...
+  // We want to use a different client if we are in preview mode:
+  const data = isPreview
+    ? await drupalPreviewer.doGraphQlRequest(
+        GET_ENTITY_AT_DRUPAL_PATH,
+        variables,
+      )
+    : await drupal.doGraphQlRequest(GET_ENTITY_AT_DRUPAL_PATH, variables);
 
   // If the data contains a RedirectResponse, we redirect to the path:
   const redirect = extractRedirectFromRouteQueryResult(data);
```
Comment (vermario) [inline next/pages/[...slug].tsx:112]: here we can switch to a different client based on the fact that we are in preview or not. Should still be readable, I think?

vermario:
I have renamed the two instances of the drupal client to drupalClientViewer and drupalClientPreviewer, that should clarify what they do.

joshua-scott (approved):
Tested according to instructions (locally and in feature environment) and everything is working as expected. Didn't spot any issues in the code either 👍

PR #215, opened 2024-05-16
NEX-154: Prevent the Drupal backend from being indexed by search engines

vermario:
We don't want the backend Drupal environment to be indexed by search engines, so in this PR we add the `noindex` metatag to all pages

To test: 

2. check that drupal pages have the `noindex` meta tag:

image

But check that frontend pages do not have the noindex tag

### DIFF
--- a/drupal/assets/robots.txt
+++ b/drupal/assets/robots.txt
@@ -1,3 +1,0 @@
-# For the Drupal backend, we want to disallow all robots:
-User-agent: *
-Disallow: /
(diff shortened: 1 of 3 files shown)


Zionknight-crypto (approved):
Changes appear according to test scenario. Approved

vermario:
> Looks good and does what it should. Is there plan for conditionally checking the environment, so everything but production is always 'noindex'?

I think that should not be necessary in our case: 

1. this change affects only the **backend** of the site, and we want to prevent indexing of it in all cases
2. for the **frontend** of the site, we can enable silta's basic auth (for all environments apart from production), which accomplishes the same thing.

phonoman:
Late to the party, but isn't `<meta name="robots" content="noindex">` the only correct way to noindex? 
I have never seen and cannot find any references to something like `<meta rel="robots" href="noindex">`

PR #216, opened 2024-05-17
NEX-154: Fix the noindex metatag

vermario:
In #215 the value for the noindex metatag was wrong, this fixes it


drupal/web/modules/custom/wunder_next/wunder_next.module
```diff
@@ -208,8 +208,8 @@ function wunder_next_page_attachments_alter(array &$attachments) {
     [
       '#tag' => 'meta',
       '#attributes' => [
-        'rel' => 'robots',
-        'href' => 'noindex',
+        'name' => 'robots',
+        'content' => 'noindex',
       ],
     ],
     'noindex',
```
phonoman, line 215: This also seems unnecessary and/or like some accidentaly extra copy-pasta?

PR #219, opened 2024-05-21
NEX-153: Add necessary authentication to graphql-codegen

joshua-scott:

Since https://github.com/wunderio/next-drupal-starterkit/pull/213, the graphql endpoint now requires authentication via an access token. This means we need to:

- Create and store an access token before running graphql-codegen script (322f4994ebda9209a982bbec9c5d1ccccda7caa4)
- Update graphql-codegen config to use that token (ee0b11a0c628ae2fdc220bd7e3570b999002be19)
- Expose the graphql schema so the vscode-graphql extension can use it. Though it previously fetched the schema itself, it now uses the file next/lib/graphql/schema.graphql that is generated by graphql-codegen (12b442358f2be88c43612fd4238cdda9601967b5)

## How to test

Checkout this PR locally, run `lando npm install`, and `lando npm run graphql-codegen`. Everything should be working as expected, including autocomplete when editing graphl queries etc.

### DIFF
--- a/next/.graphqlrc.yml
+++ b/next/.graphqlrc.yml
@@ -1,6 +1,0 @@
-# .graphqlrc.yml
-# NOTE: this needs to be http and not https:
-# Also, this should be updated with the project's lando url:
-schema: "http://next-drupal-starterkit.lndo.site/graphql"
-documents:
-  - "lib/graphql/**/*.{graphql,js,ts,jsx,tsx}"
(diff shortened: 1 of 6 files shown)


next/.graphqlrc.yml
```diff
@@ -1,6 +1,3 @@
-# .graphqlrc.yml
-# NOTE: this needs to be http and not https:
-# Also, this should be updated with the project's lando url:
-schema: "http://next-drupal-starterkit.lndo.site/graphql"
-documents:
-  - "lib/graphql/**/*.{graphql,js,ts,jsx,tsx}"
+# This file exposes the graphql schema and documents to the graphql.vscode-graphql extension
```
vermario, line 1:
```suggestion
# This file exposes the graphql schema and documents to the graphql.vscode-graphql and other compatible IDE extensions
```


I tried with the graphql extension in phpstorm and it seems to work as well.

vermario (approved):
Works as advertised, and solves a big problem :-) 

(See small suggestion)

PR #220, opened 2024-05-22
NEX-153: Remove withAuth option from Drupal client's doGraphQlRequest

joshua-scott:
Auth is now always required for GraphQL requests.


vermario:
Makes sense. Merging :)

PR #225, opened 2024-06-02
NEX-133: Add a paragraph to display an arbitrary list of articles

vermario:
## Link to ticket:

https://wunder.atlassian.net/browse/NEX-133

* BE: https://nex-133.next-drupal-starterkit.dev.wdr.io
* FE: https://nex-133-next.next-drupal-starterkit.dev.wdr.io/

This PR adds a new paragraph type where the user can select some articles and display them on the page. This is used to demonstrate entity references.

This new paragraph type can be added to `frontpage` and `page` nodes.

The migration in `wunder_democontent` adds one paragraph of this type to the frontpage, with 3 "handpicked" articles.

We are also changing a bit the way that we handle article teasers: before the typescript type was based on the article listing, but now we are using the type provided by the grapqhl fragment everywhere, which seems like a better solution.

### References to unpublished content

To test this, I have unpublished locally one of the three articles that are referenced in the paragraph.

The behaviour works correctly the unpublished content then disappears from the page, also in preview (even though the previewer role can see all published content: 

image

This is because of this setting in graphql_compose:  💎

image

## Screenshots: 

In the ui: 
image
> This widget does not allow you to alter the order of the article. In projects I have used this module https://www.drupal.org/project/multiple_selects that allows it, but I didn't want to make this feature too big... thoughts?

In the frontend (frontpage)

image

### Locally

as always: `./setup.sh -c`
* in the backend, navigate to the frontpage, see that there is a new paragraph with 3 items selected
* in the frontend, navigate to the frontpage, see that the component is rendered

### On silta

* BE: https://nex-133.next-drupal-starterkit.dev.wdr.io
* FE: https://nex-133-next.next-drupal-starterkit.dev.wdr.io/


drupal/recipes/wunder_pages/config/field.field.paragraph.accordion_item.field_content_elements.yml
```diff
@@ -13,7 +13,7 @@ dependencies:
   module:
     - entity_reference_revisions
 _core:
-  default_config_hash: gI9oa4eX3GU6bmtnilvYYWzqTIBnFP4QmP2DrvuBGlA
+  default_config_hash: 9SX29NIrIS2WBAgKvkOv-XjEtpNbr0BtMk5u4ZHc6fE
```
vermario, line 16: These changes are irrelevant, they are the product of me overriding these config files in the recipe... we could also revert this change to make this clearer... or we just ignore this :)

next/components/article/article-list-item.tsx
```diff
@@ -4,11 +4,11 @@ import { useRouter } from "next/router";
 import { useTranslation } from "next-i18next";
 import classNames from "classnames";
 
+import type { FragmentArticleTeaserFragment } from "@/lib/gql/graphql";
 import { formatDateTimestamp } from "@/lib/utils";
-import type { ArticleTeaserType } from "@/types/graphql";
 
 interface ArticleListItemProps {
-  article: ArticleTeaserType;
+  article: FragmentArticleTeaserFragment;
```
vermario, line 11: This is an example were can can use the new FragmentArticleTeaserFrament type that we get from `graphql_codegen`

joshua-scott (approved):
Tested locally and reviewed the code, working very nicely indeed :)

PR #231, opened 2024-06-13
Cleanup and simplify Drupal clients

joshua-scott:

https://feature-cleanup-drupal-clief43.next-drupal-starterkit.dev.wdr.io/

- Move to one file
- Code more DRY
- Remove unnecessary async/await and types that can be inferred
- One retry in dev mode (I have seen cases locally where first request fails but the second one succeeds)


vermario (approved):
nice improvement. works! 

Good idea to have 1 retry on dev. I will merge it.

PR #235, opened 2024-07-23
Caching Search results

micahsuomi:

Drupal: https://feature-search-caching.next-drupal-starterkit.dev.wdr.io
ssh login: `ssh www-admin@feature-search-caching-shell.next-drupal-starterkit -J www-admin@ssh.dev.wdr.io`
Next: https://feature-search-caching-next.next-drupal-starterkit.dev.wdr.io/search

- Added `queryClient` from react-query to perform search requests with `queryKey` depending if the `state.searchTerm` is changing
- When the first results are returned from ES, the search state and queryKey are added to the query client
- If the searchTerm does not change, search will return cached data. If it changes, it will perform another search request
- This also works when selecting filters

[Explain what was done and why. You can also add screenshots here if it helps.]

## How to test:

1. With the Network tab open on the console. Try to submit an search request without input query. It will return all search results with a 200 request. Try to submit again a few times, it shouldn't make the same request again but return cached data using the `queryClient`
2. Likewise, type some query on the input and press search btn or enter, and try to submit again without changing the query on the input. After the first request, it should only return cached data unless the search is changed
3. Try also with the filters. When selecting and deselecting, only the first selections will make a search request (for instance if we select the "Drupal" filter it will make one network request only. Selecting and deselecting it after that will only return cached data)
4. Compare with current production env :)

## Best practices:

<details>
<summary><h3>Accessibility:</h3></summary>
<p>
This project must support WCAG accessibility level AA <em>(edit this according to the requirements of your project)</em>. To ensure this standard is met, remember to:

- Perform automated checks using a tool such as Wave or SiteImprove.
- Test keyboard navigation: are all parts of the UI navigable using only the keyboard? Is the tab order logical? Can popups, menus etc be dismissed with the escape key?
- Test responsiveness, scaling and text reflow.
- Make sure no accessibility issues exist on either desktop or mobile views.
- If you have time, test with a screen reader such as VoiceOver (macOS), NVDA (Windows), or Orca (Linux).

Use the [Accessibility Testing Cheat Sheet](https://intra.wunder.io/info/accessibility-group/accessibility-testing-cheat-sheet) for information on how to run these tests.

</p>
</details>

### DIFF
--- a/next/pages/search.tsx
+++ b/next/pages/search.tsx
@@ -11,6 +11,7 @@ import {
   WithSearch,
 } from "@elastic/react-search-ui";
 import { SearchDriverOptions } from "@elastic/search-ui";
+import { useQueryClient } from "@tanstack/react-query";
 
 import { HeadingPage } from "@/components/heading--page";
 import { Meta } from "@/components/meta";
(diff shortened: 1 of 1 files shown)


joshua-scott (approved):
Tested locally and working exactly as described 👍

PR #236, opened 2024-07-25
NEX-101: Improve data fetching and error handling

joshua-scott:
## Link to ticket:

https://wunder.atlassian.net/browse/NEX-101

https://feature-nex-101-data-fetchi4a2.next-drupal-starterkit.dev.wdr.io

See commits for more specifics, but the main parts of this are:

- Cleanup page data fetching logic in general (getStaticProps, getCommonPageProps etc)
- Encourage faster page generation by kicking off promises as soon as possible (no more waiting until all other page data has been fetched, then `await getCommonPageProps(...)` right at the end)
- Abort the build if a page is 404 during `next build` (pages returned from getStaticPaths should always exist)
- Throw an error if the graphql query for node data is unsuccessful in getStaticProps (and provide a link to CMS logs for further debugging). This means that if there are networking or GraphQL problems, the page will not 404 (current behaviour). Instead, the page won't be updated so the last working version of the page will continue to work. (Still, if a page _should_ 404 as it's been deleted or unpublished, that will still correctly update the page to a 404)

## How to test:

1. Check CI logs build logs and feature environment, should work without errors.
2. After Silta has put the feature environments to sleep, check the site still works by resurrecting the node container but not the Drupal one.

### DIFF
--- a/next/lib/constants.ts
+++ b/next/lib/constants.ts
@@ -1,0 +1,2 @@
+export const REVALIDATE_SHORT = 10; // 10 seconds
+export const REVALIDATE_LONG = 60 * 10; // 10 minutes
--- a/next/lib/contexts/language-links-context.tsx
+++ b/next/lib/contexts/language-links-context.tsx
@@ -37,10 +37,10 @@ export const getStandardLanguageLinks = () =>
  */
 export function createLanguageLinksForNextOnlyPage(
   path: string,
-  context: GetStaticPropsContext,
+  locales: GetStaticPropsContext["locales"],
 ): LanguageLinks {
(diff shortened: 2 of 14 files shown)


vermario:
I think this is a nice improvement and we can merge it.


I suggest we open a followup ticket regarding what happens when you unpublish a translation of the homepage.


at the moment though we have a bigger problem related to https://github.com/wunderio/next-drupal-starterkit/pull/205 , due to what I think is a bug in the graphql_compose module https://www.drupal.org/project/graphql_compose/issues/3464624#comment-15705017

vermario (approved):

PR #243, opened 2024-09-06
NEX-127: Switch from pages router to app router

vermario:
## App router switch

A long list of changes in this PR. Major kudos to @jekku123 that has done 99% of the work for this! 🥇 

* Switch to App router from Pages router
* Update to drupal 10.3.9
* Update contrib modules
* Update to latest version of next-drupal (beta2) both Drupal module and npm package
* Switch to Auth.js for authentication
* Changes to drupalClient to extend NextDrupalBase
* Switch to next-intl
* Refinements for the redis integration
* Implementation of Draft Mode
* Directory refactoring in the /next subfolder
* Revamped search functionality using searchkit
* Added dark mode


vermario:
@jekku123 hi! I tried testing this a bit, and I think we have an issue in the caching setup:
image

Basically the changes saved to the page do not show up unless you are in draft mode.

Either next.js or redis are not getting invalidated, it seems.

Zionknight-crypto:
When going to edit Frontpage I see this preview and when I click that link it renders the Frontpage so is the preview feature creating a new page and returning 404 by default? Anyone else seen this? Actually this seems to be the case

Screenshot 2024-11-14 at 13 59 06

I could not update Frontpage text content and Articles could not be fetched.

Screenshot 2024-11-14 at 14 06 25

Search also does not work. 

Could others confirm whether or not this is the case for them as well so we can check as a team in parallel?

vermario:
> When going to edit Frontpage I see this preview and when I click that link it renders the Frontpage so is the preview feature creating a new page and returning 404 by default? Anyone else seen this? Actually this seems to be the case
> Screenshot 2024-11-14 at 13 59 06
> 
> I could not update Frontpage text content and Articles could not be fetched.
> Screenshot 2024-11-14 at 14 06 25
> 
> Search also does not work.
> 
> Could others confirm whether or not this is the case for them as well so we can check as a team in parallel?

Can't reproduce :-) 

Showing Frontpage:

image

Search: 

image

Articles:

image

Zionknight-crypto:
@vermario mmm that is odd indeed, I need to re-setup again from start.

vermario:
Something that happens when switching branches is that you might have some old generated code laying around, since this new feature changes the whole structure, and the files generated by the graphl-codegen are now in a new location. 
so I suggest doing this: 
1. `git clean -f` <- should get rid of leftover untracked files
2. `./setup-lando.sh -c`  will rebuild lando and reinstall drupal and should get you to the initial situation.

Zionknight-crypto (approved):
Making a pre-approval just in case.

joshua-scott (approved):
Don't have enough time to go through every change here, but I tested this by running the lando setup command locally. Everything ran successfully and looks to be working great!

+1 for promoting the app router as default soon, and preserving the pages router under a `pages-router` branch

PR #244, opened 2024-09-08
Nex 127 search

vermario:
DRAFT, adding search to app router


jekku123 (approved):
Works great! Nice to have this finally with the app router :)

PR #250, opened 2024-09-27
NEX-172: Fix preview of previous revisions of nodes

vermario:
## Link to ticket:

https://wunder.atlassian.net/browse/NEX-172 

We were not correctly passing the node id when doing the graphql query to get the entity at a certain revision. This PR fixes that.

## How to test:

1. edit a piece of content
2. a revision is created
3. click on the "revisions" tab
4. click on the link of the previous revision
5. the content is shown at that revision

### DIFF
--- a/next/pages/[...slug].tsx
+++ b/next/pages/[...slug].tsx
@@ -191,8 +191,8 @@ export const getStaticProps: GetStaticProps<PageProps> = async ({
     typeof previewData.resourceVersion === "string" &&
     previewData.resourceVersion !== "rel:latest-version"
   ) {
-    const [nodeId, revisionId] = previewData.resourceVersion.split(":");
-    const path = `/node/${nodeId}/revisions/${revisionId}/view`;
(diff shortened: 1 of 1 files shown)


next/pages/[...slug].tsx
```diff
@@ -191,8 +191,8 @@ export const getStaticProps: GetStaticProps<PageProps> = async ({
     typeof previewData.resourceVersion === "string" &&
     previewData.resourceVersion !== "rel:latest-version"
   ) {
-    const [nodeId, revisionId] = previewData.resourceVersion.split(":");
-    const path = `/node/${nodeId}/revisions/${revisionId}/view`;
+    const [_, revisionId] = previewData.resourceVersion.split(":");
+    const path = `/node/${node.id}/revisions/${revisionId}/view`;
```
Comment (vermario) [inline next/pages/[...slug].tsx:195]: the resourceVersion property looks like this: 

"id:20" for a revision with id 20. 

So the existing code was just wrong :)

jekku123 (approved):
Tested and approved!

PR #254, opened 2024-10-15
Tell Lando node service to use the openssl CA

vermario:
* This has become necessary in newer versions of lando to be able to access the drupal service from within lando using https.

(See https://github.com/lando/core/issues/229 )


MarttiR (approved):

vermario:
@Tromppa @Zionknight-crypto @joshua-scott I recommend to add this same change to your projects. :)

PR #258, opened 2024-10-18
NEX-175: Add the possibility of running npx commands in lando

vermario:
This branch adds the npx command to lando.

To test:

1. switch to this branch
2. if you were working with an app router branch before (it will delete untracked file, so be careful?) `git clean -d -f`
3. `./setup-lando.sh -c`
4. stop the server with control-c
5. cd next (the npx command will be executed in the directory you are in
6. for example: `lando npx shadcn@latest add badge`

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -26,6 +26,10 @@ tooling:
     service: node
     dir: /app/next
 
+  npx:
+    command: "npx"
+    service: node
+
   # This command will stop (kill) any running node operations
   # in the node container in lando. You can use it instead of the
   # usual ctrl-c keyboard combination.
@@ -154,6 +158,8 @@ services:
         # We instruct node to use the Lando CA certificate for HTTPS connections:
         NODE_EXTRA_CA_CERTS: /lando/certs/LandoCA.crt
     build:
+      # This is needed to be able to run lando npx commands:
+      - "mkdir -p /var/www/.npm-global/lib"
       - "cd next && npm ci"
     scanner: false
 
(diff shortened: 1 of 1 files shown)


.lando.yml
```diff
@@ -154,6 +158,8 @@ services:
         # We instruct node to use the Lando CA certificate for HTTPS connections:
         NODE_EXTRA_CA_CERTS: /lando/certs/LandoCA.crt
     build:
+      # This is needed to be able to run lando npx commands:
+      - "mkdir -p /var/www/.npm-global/lib"
```
vermario, line 162: for the `node`user inside the docker container /var/www is the home folder, the directory needs to be there for npx to work, apparently.

tormi:
```yml
tooling:
  npx:
    service: node
```

doesn't work?

jekku123: Seems to work both ways.

Lets add 
```
dir: /app/next
```
there too.

vermario: @tormi without this line: 

1. lando destroy
2. lando start
3. 
image

jekku123 (approved):
Awesome, makes Shadcn'izing much easier :)

.lando.yml
tormi: Ok, thanks for the explanation.

tormi: I suggest installing `npm` globally instead. This would be a cleaner solution:

```yml
services:
  node:
    type: node:20
    globals:
      npm: latest
```

vermario: nice! I tested it and it works, let's switch to that.

Committed in faefbacaa5db0330953ef597cdf95438f39a56bd

PR #259, opened 2024-10-21
NEX-170: Make example content even more clearly example content

vermario:
In this branch :

1. we make the test content even more clearly test content by removing the AI generated reasoning about next.js and drupal
1. we reduce the amount of pages (content type page) to the minimum needed to show the three levels in the menu


Zionknight-crypto (approved):
Works for me ;)

PR #273, opened 2025-01-31
NEX-183: Improve preview mechanism by adding graphql_compose_preview 

vermario:
## Link to ticket:

https://wunder.atlassian.net/browse/NEX-183

* Add the graphql_compose_preview so we can preview entities with the data the user has input into the form, 
* setup all content types to use preview, configure the field formatter to point to the frontend adding to the recipe
* add a page to the frontend to display the preview, to be embedded in the iframe

Here's how it looks like: 

https://github.com/user-attachments/assets/222eeda9-8556-4e00-b64e-c652e0d301f0

### To test

* clone
* checkout this branch
* ./setup-ddev.sh or ./setup-lando.sh
* open an existing node, edit the content, click on the preview button, you should see it in the iframe
* create a new node, add some content to the form, click on preview

### TODO: 

Note: We added a patch to be able to preview a new node: https://www.drupal.org/project/graphql_compose/issues/3503977  . When the next version of the module is released, we need to remember to remove it.


drupal/patches/recipe-10.2.x.patch
vermario, line 1: This patch was not used anymore.

next/src/lib/drupal/get-node.ts
vermario, line 1: Here we export also this function so we can use it directly and bypass caching

Tromppa (approved):
This is very nice!

jekku123 (approved):
Tested and happy!

drupal/recipes/wunder_pages/config/core.entity_view_display.node.page.default.yml
```diff
@@ -8,32 +8,31 @@ dependencies:
     - field.field.node.page.field_image
     - node.type.page
   module:
-    - entity_reference_revisions
+    - graphql_compose_preview4
```
phonoman, line 11: @vermario `4` looks like a typo?

vermario: yes it is!

fixed in https://github.com/wunderio/next-drupal-starterkit/pull/275/commits/841384342453e61c3496f717a6b27731682bb224 thanks!

PR #276, opened 2025-02-11
NEX-84: Support forward revisions when displaying a revision in the node page in drupal

vermario:
This PR adds:

1. enabling of the workflows and content moderation modules
2. add patch at https://www.drupal.org/project/graphql_compose/issues/3462517 for graphl_compose to add the possibility of adding the revision to a route query
3. adapting the frontend code to pass the version in the expected way

## How it looks

https://github.com/user-attachments/assets/ea680c08-20b0-4631-b7ea-4213279b0bb8

## How to test

1. `./setup-lando.sh -c`
2. log into the backend, edit a "page", create a draft revision, preview it at node/id
3. create a new page, save it in as a draft, you should be able to see it at node/id


Tromppa (approved):
Nice that you got this working also!

PR #278, opened 2025-02-12
NEX-186: Update core and contrib modules, upgrade to php8.3, sync with drupal-project, expand tests

vermario:

Time to update core and the contrib modules we use. 

* Update to Drupal core 10.4.2
* Update the webform module to 6.2.9
* Update the subrequests module to 3.0.12
* Update the redirect module to 1.11
* Update the metatag module to 2.10
* Update the graphql module to 4.10 and typed_data to 2.1.1
* Update the decoupled_router  module to the 2.0.5 version
* Update the require_login module to the 3.0.7 version
* Update the graphl_compose module to the 2.3 version

In this PR we also expand the cypress tests to cover the menu and the dashboard.

Bonus: NEX-186: Fix translation settings for the file attachments paragraph in recipe, it was set as untranslatable so it showed up always in swedish

## How to test:

As always, start from scratch with 

`./setup-lando.sh -c`

Then, run the cypress tests, by opening a new terminal and executing

`lando npm run cypress:run`

Also, use the site. Areas potentially impacted are:

* forcing login
* graphql queries
* preview
* webrforms
* redirects

The updates page should look like this: 
image

So all green, apart from webform rest that required the beta version of webform to be updated as well, so it can wait.


.lando.yml
```diff
@@ -76,6 +76,7 @@ services:
   database:
     # This is the port you can use to connect to mysql from your host.
     portforward: 34567
+    type: "mariadb:10.6.18"
```
vermario, line 79: newer mariadb version. Syncs to what we have at drupal-project.

.lando.yml
```diff
@@ -213,10 +211,24 @@ services:
...
+        - .lando/data-mailpit:/data
+        ports:
+          - "8025"  # Web UI
+          - "1025"  # SMTP
+        environment:
+          MP_MAX_MESSAGES: 5000
+          MP_DATA_FILE: /data/mailpit.db
+          MP_SMTP_AUTH_ACCEPT_ANY: 1
+          MP_SMTP_AUTH_ALLOW_INSECURE: 1
+        command: "/mailpit"
```
vermario, line 231: mailhog is not supported anymore. Mailpit is, and it also integrates nicely with cypress https://www.npmjs.com/package/cypress-mailpit

PR #279, opened 2025-02-14
NEX-187: Fix issues with latest version tab and translated entities

vermario:
After merging #276 , we noticed that in multilingual content, the "latest revision" tab would always show the node in the default language, even if a translation existed.

## The issue

https://github.com/user-attachments/assets/999083b7-3481-496c-8776-7c55f3ddf975

## How we fix it

In this PR we fix this issue by leveraging the `_next_next_site_preview_alter` hook, checking if the entity has a translation for the active language of the site, and switching to the translation if so.


phonoman (approved):
Works great!

PR #289, opened 2025-04-12
NEX-195: Allow the frontend to access the backend via internal url in runtime

vermario:
This PR introduces the concept of an internal url to access drupal from the frontend.

The backend needs to be accessible by a public URL for these reasons:
1. the editors need to log into the site and do their content editing work
2. in our example, there is a flow where the user asks to register on the frontend, gets an email with the link to go to drupal and set the password, gets redirected back
3. the build of the site is done in our case by CircleCI and it's not feasible to allow access to the internal url because the circle ci agent does not have a predictable IP.

But, it would be useful for the node.js process running in next.js to access the backend also with an **internal URL**. In SILTA, this means that the data flows inside the same cluster, and does not involve our CDN.

One thing that we had to take into consideration was the fact that we are using `next-image`, and both the internal and external urls need to be allowed as source for the images.

I have added the same concept to the lando and ddev setups, because hopefully this would allow us to catch subtle bugs in case we are using the wrong internal/external url for drupal there.

Summary of the changes:

### Configuration Updates:
* [`.circleci/config.yml`](diffhunk://#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R126-R128): Added environment variables `DRUPAL_BASE_URL_INTERNAL` and `DRUPAL_BASE_URL_INTERNAL_IMAGES` to handle internal URLs and ensure they are used within CircleCI. [[1]](diffhunk://#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R126-R128) [[2]](diffhunk://#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R151-R156) [[3]](diffhunk://#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R166-R170)
* [`.ddev/config.yaml`](diffhunk://#diff-ca25b57a32546d43021f7de9e5374004e498a3cc2ec9b4be3c820daa1e896e57R39-R40): Added `DRUPAL_BASE_URL_INTERNAL` and `DRUPAL_BASE_URL_INTERNAL_IMAGES` for local development.
* [`.lando.yml`](diffhunk://#diff-307b4548b91e997f23c35d90ed5e9a044e156b90425c9923d6b85d7dc11e67e2R143-R144): Added `DRUPAL_BASE_URL_INTERNAL` and `DRUPAL_BASE_URL_INTERNAL_IMAGES` for Lando environments.
* [`silta/silta-next.yml`](diffhunk://#diff-e0668b5e0f7cb8d1dac9faa565be35910f0a6c176eb658aead285eaed68329efR21-R22): Added `DRUPAL_BASE_URL_INTERNAL` and `DRUPAL_BASE_URL_INTERNAL_IMAGES` for Silta environments.

### Code Updates:
* [`next/next.config.mjs`](diffhunk://#diff-b3ada88456dc03ef295134b2cb2f5eca433ba810c81559e3904816a6d6ab07a8L8-R26): Updated the image hostname handling to use both external and internal Drupal URLs, ensuring valid URLs are provided. [[1]](diffhunk://#diff-b3ada88456dc03ef295134b2cb2f5eca433ba810c81559e3904816a6d6ab07a8L8-R26) [[2]](diffhunk://#diff-b3ada88456dc03ef295134b2cb2f5eca433ba810c81559e3904816a6d6ab07a8L28-R50)
* [`next/src/app/api/health/route.ts`](diffhunk://#diff-bc8bb7f9074889ff91d4fb0985c03e0a6729006572e625c125bb7fb465dbba4fL9-R9): Changed the health check endpoint to use the internal Drupal URL.
* [`next/src/auth.ts`](diffhunk://#diff-d489c058ed586206f4e44093feaad32ad3c037fde31246cc6db7920c45edaffeL47-R47): Updated OAuth token fetching to use the internal Drupal URL. [[1]](diffhunk://#diff-d489c058ed586206f4e44093feaad32ad3c037fde31246cc6db7920c45edaffeL47-R47) [[2]](diffhunk://#diff-d489c058ed586206f4e44093feaad32ad3c037fde31246cc6db7920c45edaffeL118-R118)
* [`next/src/components/formatted-text.tsx`](diffhunk://#diff-909dc080ccef96c47e4360f07105f8dbfb0c1dc35435c301a18e80cb37276958L37-R37): Modified image source URLs to use the internal Drupal URL.
* [`next/src/env.ts`](diffhunk://#diff-1f8511ce94c4bcac8eaec4e358ce3ee874c4b8e1b6979a9de3bba62205e0e0d9R13-R14): Added internal Drupal URL environment variables to the environment configuration. [[1]](diffhunk://#diff-1f8511ce94c4bcac8eaec4e358ce3ee874c4b8e1b6979a9de3bba62205e0e0d9R13-R14) [[2]](diffhunk://#diff-1f8511ce94c4bcac8eaec4e358ce3ee874c4b8e1b6979a9de3bba62205e0e0d9R32-R34)
* [`next/src/lib/drupal/drupal-client.ts`](diffhunk://#diff-507fa61ba098957b12722be69aa79ded893b916d765928fa30076e72021db294L46-R46): Updated the GraphQL client to use the internal Drupal URL.


silta/silta-next.yml
```diff
@@ -18,6 +18,7 @@ services:
       DRUPAL_REVALIDATE_SECRET: ${DRUPAL_REVALIDATE_SECRET}
       AUTH_SECRET: ${AUTH_SECRET}
       NEXT_PUBLIC_DRUPAL_BASE_URL: https://<|DRUPAL_DOMAIN|>
+      DRUPAL_BASE_URL_INTERNAL: https://<|DRUPAL_DOMAIN|>
```
vermario, line 21: @Jancis how do we get here the internal url for drupal?

Jancis: it should be `<release-name>-drupal`

vermario: is there a risk for the release name to be too long because of the branch name and this failing for that?

vermario: so we should add another task after this one: 

https://github.com/wunderio/next-drupal-starterkit/blob/0066792accc71ea8ca79f32dcc47314b8fb7e19a/.circleci/config.yml#L307 


Where we again replace the value of the env var DRUPAL_BASE_URL_INTERNAL to this new value...

.circleci/config.yml
```diff
@@ -123,6 +123,9 @@ commands:
 
             drupal_release_name=`silta ci release name`
             next_release_name=`silta ci release name --release-suffix next`
+            # This is the name that will be used in the internal URL, used in
+            # Node containers to call the Drupal backend.
+            drupal_release_name_for_internal_use=${drupal_release_name}-drupal
```
MarttiR, line 128: I'm wondering whether this works with long branchnames that get truncated and hashed. `silta ci release name --release-suffix drupal` might work better? One way to test this would be to just create a long branchname and push it to see if it works after deployment.

Side note: Not sure if the manual truncation in this script is still needed as `silta ci release name` will already truncate long names, including the release suffixes.

Jancis: It looks like it would work, but it's incorrect usage. The reason - you use `release-suffix: next` and `release-suffix: storybook` but for drupal there is no release suffix, just release name. And `-drupal` is a "load balancer hostname" (service)

vermario: Thanks for clarifying this!

hkirsman:
I liked also this explanation from Slack thread:

```
Main issue:

If the frontend uses the fastly-protected public url for drupal, we unnecessarily generate fastly traffic (since drupal also runs behind fastly, as for our normal procedures), and we might incur in issues with rate limiting (since the frontend is seen as an individual "client" doing a lot of calls, and it might be blocked,
```

Zionknight-crypto (approved):
Looks good to me, not the greatest at configs but this looks clear enough. Where exactly do we retrieve and see these internal urls? It would be great to document this with a short readme eventually.

vermario:
I found this issue in tests: 
image

basically the email link is generated using the internal url for drupal this way... will have to fix

vermario:
> I found this issue in tests: image
> 
> basically the email link is generated using the internal url for drupal this way... will have to fix

Fixed in https://github.com/wunderio/next-drupal-starterkit/pull/289/commits/b5cfb18fd623e35686944db081291118dbe5f454

.lando.yml
```diff
@@ -140,6 +140,8 @@ services:
         DRUPAL_CLIENT_VIEWER_ID: drupal-client-viewer-id
         DRUPAL_CLIENT_VIEWER_SECRET: drupal_client_viewer_secret_not_secure_used_only_locally
         NEXT_PUBLIC_DRUPAL_BASE_URL: https://next-drupal-starterkit.lndo.site
+        DRUPAL_BASE_URL_INTERNAL: http://appserver_nginx.nextdrupalstarterkit.internal
+        DRUPAL_BASE_URL_INTERNAL_IMAGES: http://appserver_nginx.nextdrupalstarterkit.internal
```
joshua-scott, line 144: In lando also just `http://appserver_nginx` should work, I don't think the `.nextdrupalstarterkit.internal` part is needed. I suppose it's one less thing that would need to be changed later

next/next.config.mjs
```diff
@@ -25,8 +41,13 @@ const nextConfig = {
...
       {
-        protocol: "https",
-        hostname: imageHostname,
+        protocol: imageHostnameExternal[0] === "https" ? "https" : "http",
+        hostname: imageHostnameExternal[1],
+        pathname: "**",
+      },
+      {
+        protocol: imageHostnameInternal[0] === "https" ? "https" : "http",
+        hostname: imageHostnameInternal[1],
```
joshua-scott, line 50: In another project, we kept logic for this in one place like this:

```ts
remotePatterns: [
    process.env.DRUPAL_BASE_URL_INTERNAL_IMAGES,
    process.env.NEXT_PUBLIC_DRUPAL_BASE_URL,
  ].map((url = "") => {
    const [protocol, hostname] = url.split("://");
    if (!hostname || (protocol !== "https" && protocol !== "http")) {
      throw new Error(`Invalid images URL "${url}" in next.config.ts`);
    }
    return {
      protocol,
      hostname,
      pathname: "**",
    };
  }),
```

next/src/components/formatted-text.tsx
```diff
@@ -34,7 +34,7 @@ const options: HTMLReactParserOptions = {
         if (isRelative(src)) {
           return (
             <Image
-              src={`${env.NEXT_PUBLIC_DRUPAL_BASE_URL}${src}`}
+              src={`${env.DRUPAL_BASE_URL_INTERNAL}${src}`}
```
joshua-scott, line 37: I think here we should actually use `NEXT_PUBLIC_DRUPAL_BASE_URL`, or are we 100% sure this component will only ever load from the server?

In another project (pages router but otherwise very similar) where this line in FormattedText caused an error "Attempted to access a server-side environmental variable on the client", so had to change this one

joshua-scott (commented):
This approach here is now working nicely on a live site too! A few comments from that implementation to consider, then this should be good to go

.lando.yml
joshua-scott: Done in 339c79a

joshua-scott (approved):

PR #290, opened 2025-04-29
NEX-179: Move the redirect logic to middleware

vermario:
Work in progress:

* console.log messages left on purpose
* one call to drupal for each page + prefetch links -> not great

### DIFF
--- a/next/src/app/[locale]/(dynamic)/[...slug]/page.tsx
+++ b/next/src/app/[locale]/(dynamic)/[...slug]/page.tsx
@@ -1,6 +1,6 @@
 import { Metadata } from "next";
 import { draftMode } from "next/headers";
-import { notFound, permanentRedirect, redirect } from "next/navigation";
+import { notFound } from "next/navigation";
 import { getDraftData } from "next-drupal/draft";
 import { setRequestLocale } from "next-intl/server";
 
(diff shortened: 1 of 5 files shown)


next/src/app/[locale]/(dynamic)/[...slug]/page.tsx
Comment (vermario) [inline next/src/app/[locale]/(dynamic)/[...slug]/page.tsx:1]: Basically here the idea is that once we get here we dont' need to check if there's a redirect, because that is handled by the middleware.

next/src/lib/drupal/get-node-nocache.ts
vermario, line 1: I had to move this function to its own file, because we could not import an export from this file in middleware (the neshka/cache package causes webpack errors there)

next/src/middleware.ts
```diff
@@ -60,6 +136,12 @@ const authMiddleware = (request: NextRequest, ctx: AppRouteHandlerFnContext) =>
       }
     }
 
+    // Check if there are redirects from Drupal for this request:
+    const redirectResponse = await handleDrupalRedirects(request);
+    if (redirectResponse) {
```
vermario, line 141: I added this here, after the checks for auth pages, and before the middleware for next-intl 

(not sure this is great)

next/src/app/[locale]/(dynamic)/[...slug]/page.tsx
```diff
@@ -77,11 +60,6 @@ export default async function NodePage({
     notFound();
   }
 
-  // If the node is a frontpage, redirect to the frontpage:
-  if (!isDraftMode && node.__typename === "NodeFrontpage") {
-    redirect(`/${locale}`);
-  }
-
```
Comment (vermario) [inline next/src/app/[locale]/(dynamic)/[...slug]/page.tsx:84]: ah, also this we would need to handle in middleware maybe

